PR #48, opened 2023-02-01
feat: add support for mailbox / relay

jrriehl:
New example `11-mailbox-agents` shows how to use the new mailbox server. The current default is to use one that is running locally.


src/uagents/relay.py
```diff
@@ -0,0 +1,94 @@
...
+                else:
+                    logging.exception(
+                        f"Failed to retrieve messages from inbox: {(await resp.text())}"
+                    )
+
+            if success and len(items) > 0:
+                async with session.delete(
+                    mailbox_url,
+                    headers={"Authorization": f"token {self._access_token}"},
+                ) as resp:
```
ejfitzgerald, line 54: Deleting everything in the mailbox is a little dangerous because a subsequent message might have come in. I suggest that you iterate over the message ids and delete  them individually

examples/11-mailbox-agents/agent1.py
```diff
@@ -0,0 +1,38 @@
...
+    name="alice",
+    port=8010,
+    seed="agent1 secret phrase",
+    use_relay=True,
+    resolve=RulesBasedResolver(
+        {
+            AGENT1_ADDRESS: f"{RELAY_SERVER_URL}/v1/submit",
+            AGENT2_ADDRESS: f"{RELAY_SERVER_URL}/v1/submit",
+        }
+    ),
```
ejfitzgerald, line 23: My thought process for the configuration something like this:

```python
Agent(
    ...
    mailbox='<api key>',
)
```

or for the longer version:

```python
Agent(
    ...
    mailbox='<api key>@<mailbox base url>',
)
```

or the full version

```python
Agent(
    ...
    mailbox={
        'base_url': '<mail box base url>',
        'api_key': '<api key>',
    }
)
```

jrriehl: OK, `api_key` is not implemeted yet, right? Or did I miss that?

ejfitzgerald (approved):

PR #53, opened 2023-02-10
feat: add wallet messaging

jrriehl:
Adds wallet messaging as an optional feature. This means it needs to be installed with:
- `pip install uagents[wallet]`
- `pip install uagents[all]`
- `poetry install -E wallet`
- `poetry install -E all`

Adds `on_wallet_message` decorator and uses babble client to send and retrieve messages from wallet messaging server.

Example `15-wallet-messaging` demonstrates the new feature.


.pylintrc
```diff
@@ -7,5 +7,7 @@ disable=missing-module-docstring,
         too-many-instance-attributes,
         duplicate-code,
         too-many-arguments,
-        logging-fstring-interpolation
+        logging-fstring-interpolation,
+        import-outside-toplevel,
+        too-many-statements
```
jrriehl, line 12: :frowning_face:

pyproject.toml
```diff
@@ -29,6 +29,9 @@ tortoise-orm = "^0.19.2"
 [tool.poetry.group.geo.dependencies]
 geopy = "^2.3.0"
 
+[tool.poetry.group.wallet.dependencies]
+babble = {path = "../babble"}
```
jrriehl, line 33: Need to have this locally until we make the repo public.

src/uagents/agent.py
```diff
@@ -159,7 +170,7 @@ def sign_digest(self, digest: bytes) -> str:
 
     def sign_registration(self) -> str:
         return self._identity.sign_registration(
-            self._reg_contract.address, self.get_registration_sequence()
+            self._reg_contract.address, self._get_registration_sequence()
```
jrriehl, line 173: Made this private because of pylint complaint, but also because I think it should be.

src/uagents/wallet_messaging.py
```diff
@@ -0,0 +1,59 @@
...
+        while True:
+            self._logger.info("Checking wallet messages...")
+            for msg in self._client.receive():
+                await self._message_queue.put(msg)
+            await asyncio.sleep(self._poll_interval)
+
+    async def process_message_queue(self, ctx: Context):
+        while True:
+            msg = await self._message_queue.get()
+            for handler in self._message_handlers:
```
jrriehl, line 58: For now, we just run all the message handlers on each incoming message. We can certainly add some differentiating handlers in the future.

Archento:
I can't poetry install `frozenlist` and `pyyaml`. This may be due to my pipx venv using Python3.12
Are these mandatory and is there a way to test python compatibility issues?

jrriehl:
> I can't poetry install `frozenlist` and `pyyaml`. This may be due to my pipx venv using Python3.12 Are these mandatory and is there a way to test python compatibility issues?

We certainly don't claim to support Python 3.12 yet. Can you try with an earlier version for now and then we can raise an issue to fix this?

Archento:
> We certainly don't claim to support Python 3.12 yet. Can you try with an earlier version for now and then we can raise an issue to fix this?

When forcing 3.11.6 with all extensions installed I still get errors but this time from coincurve (didn't we got rid of that a while ago?)

Archento:
Following the instructions on their page ([docs](https://ofek.dev/coincurve/install/)) for MacOS:
```sh
xcode-select --install
brew install autoconf automake libffi libtool pkg-config python
```
I was able to install `coincurve` and since I already had the rest - it seems that one or all of 'libffi, libtool and pkg-config' were missing.

python/src/uagents/wallet_messaging.py
```diff
@@ -0,0 +1,82 @@
...
+                return func(*args, **kwargs)
+
+            self._message_handlers.append(func)
+
+            return handler
+
+        return decorator_on_message
+
+    async def send(self, destination: str, msg: WalletMessage, msg_type: int = 1):
+        self._client.send(destination, msg, msg_type)
```
Archento, line 60: The client `.send()` method requires a string but the WalletMessage is an object composed of multiple fields.
Did you mean `msg.text`? Based on the example I would assume we change the `msg` attribute for the `WalletMessagingClient` to `str` as well.

python/src/uagents/context.py
```diff
@@ -554,3 +558,11 @@ async def send_raw_exchange_envelope(
...
             endpoint="",
         )
+
+    async def send_wallet_message(
+        self,
+        destination: str,
+        text: str,
+        msg_type: int = 1,
+    ):
+        await self._wallet_messaging_client.send(destination, text, msg_type)
```
Archento, line 568: We should check if the wallet client is None and/or add other checks as this will be the main interface for the developer.

python/src/uagents/agent.py
```diff
@@ -156,6 +156,8 @@ def __init__(
         agentverse: Optional[Union[str, Dict[str, str]]] = None,
         mailbox: Optional[Union[str, Dict[str, str]]] = None,
         resolve: Optional[Resolver] = None,
+        enable_wallet_messaging: Optional[Union[bool, Dict[str, str]]] = False,
```
Archento, line 159: missing doc-string description for both.

python/src/uagents/agent.py
```diff
@@ -294,12 +300,51 @@ def _initialize_wallet_and_identity(self, seed, name):
...
+                        seed, LEDGER_PREFIX, wallet_key_derivation_index
+                    )
+                ),
                 prefix=LEDGER_PREFIX,
             )
         if name is None:
             self._name = self.address[0:16]
 
+    def initialize_wallet_messaging(
+        self, enable_wallet_messaging: Union[bool, Dict[str, str]]
```
Archento, line 314: What's the idea behind `Dict[str, str]` -> what values do we expect to come up here?
{"chain_id": "???"} 

Also, as with the `endpoint` and other definitions: the myriad ways to instantiate an agent are not well documented. We need to keep that in mind when providing such flexible interfaces.

python/src/uagents/wallet_messaging.py
```diff
@@ -0,0 +1,82 @@
...
+            signed_bytes,
+            BabbleIdentity(wallet.signer().private_key_bytes),
+            chain_id,
+        )
+        self._poll_interval = WALLET_MESSAGING_POLL_INTERVAL_SECONDS
+        self._logger = logger or get_logger("wallet_messaging")
+        self._message_queue = asyncio.Queue()
+        self._message_handlers: List[WalletMessageCallback] = []
+
+    def on_message(
```
Archento, line 45: As there is no qualifier for the handler (e.g. like a model on which to act), a developer can create multiple handlers which will all trigger at the same time whenever a wallet message is incoming.

Just raising this here as we don't prevent a dev from doing stupid things we may want to either:
- append the function to the existing handler,
- overwrite the existing handler, or
- prevent the handler from being overwritten, thus forcing the dev to structure their code differently.

Archento (changes requested):
I have just a couple of remarks but all in all it's very good!

It works as intended (when the correct environment is present) and I am curious to see this in action in more complex agent scenarios. The only downside is that this implementation does not seamlessly fit into the existing uagents codebase or looking at it from a different angle it's the perfect example for an add-on/extra module.

python/src/uagents/agent.py
jrriehl: Yes, I believe this is primarily to allow configuring the chain id, as the messaging server needs to know this. But since only our own testnet and mainnet are allowed, I'm not sure it makes sense to expose this in the API.

jrriehl: OK, I think this is because the chain id used for wallet messaging can be mainnet even while the agent is only transacting on the testnet, so we needed a way to configure this separately.

jrriehl: Added this explanation to the docstring.

Archento: I think it would be sensible to bind the messaging to the network the agent is transacting on to somewhat force developers when we switch to main net.
At least I don't see a reason for us to support agents running on one net and sending messages on the other.

python/src/uagents/context.py
```diff
@@ -554,3 +558,11 @@ async def send_raw_exchange_envelope(
             destination=destination,
             endpoint="",
         )
+
+    async def send_wallet_message(
+        self,
+        destination: str,
+        text: str,
+        msg_type: int = 1,
```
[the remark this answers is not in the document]
jrriehl: From the `babble` package:
```
...
    "type": msg_type,  # 1 for text message, 2 for transaction data
...
```
This would be better as an `int, Enum` type, but I think that should go into the `babble` package rather than here.

python/src/uagents/context.py
jrriehl: Good point, fixed.

python/src/uagents/wallet_messaging.py
jrriehl: Yes, I made the initial quick decision to allow multiple handlers to be registered for wallet messages and haven't really revisited this. Not sure I'm convinced that these options are better, but it's probably worth creating an issue to come back to this.

python/src/uagents/wallet_messaging.py
jrriehl: Good catch, actually the type hint in the function header is wrong and `WalletMessage` is not needed. Fixed.

python/src/uagents/wallet_messaging.py
jrriehl: https://github.com/fetchai/uAgents/issues/206

python/src/uagents/agent.py
jrriehl: At the moment, you can set the chain in the Fetch wallet but our agents are only running on the Dorado testnet, so we needed them to be able to communicate on mainnet even though they register on testnet. I agree that we will need to revisit this when migrating to mainnet, though.

Archento (approved):
All good from my side now!

PR #56, opened 2023-02-15
feat: add websocket support for mailbox

jrriehl:
Updates example 11 to use websockets by default. The protocol (ws(s) vs http(s)) is automatically parsed from the config entry. For example:
```python
agent = Agent(
    name="bob",
    seed="bob secret phrase",
    mailbox="my_api_key@ws://127.0.0.1:8000",
)
```


src/uagents/mailbox.py
```diff
@@ -14,32 +16,82 @@
...
+            return
+
+        await dispatcher.dispatch(
+            env.sender,
+            env.target,
+            env.protocol,
+            env.decode_payload(),
+        )
+
+        # delete envelope on server
```
jrriehl, line 79: We could consider putting these in a deletion queue instead to increase reception throughput.

src/uagents/setup.py
```diff
@@ -16,3 +19,18 @@ def fund_agent_if_low(agent_address: str):
         LOGGER.info("Adding funds to agent...")
         faucet.get_wealth(agent_address)
         LOGGER.info("Adding funds to agent...complete")
+
+
+def register_agent_with_mailbox(agent: Agent, email: str):
```
jrriehl, line 24: This will most likely need to change with uagents-relay auth updates.

src/uagents/mailbox.py
jrriehl: Done.

ejfitzgerald (approved):

PR #59, opened 2023-02-24
feat: simplify cosmpy txs and add example

jrriehl:
I think this simplified tx flow is all we need at the moment. I applied this to the registration tx and also added a send tokens example.

### DIFF
--- a/src/uagents/network.py
+++ b/src/uagents/network.py
@@ -1,6 +1,17 @@
+import asyncio
+from datetime import datetime, timedelta
+from typing import Optional
+
 from cosmpy.aerial.contract import LedgerContract
-from cosmpy.aerial.client import LedgerClient, NetworkConfig
+from cosmpy.aerial.client import (
+    LedgerClient,
+    NetworkConfig,
+    DEFAULT_QUERY_INTERVAL_SECS,
+    DEFAULT_QUERY_TIMEOUT_SECS,
+)
+from cosmpy.aerial.exceptions import NotFoundError, QueryTimeoutError
 from cosmpy.aerial.faucet import FaucetApi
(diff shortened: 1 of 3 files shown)


ejfitzgerald (approved):
Yeah makes a lot of sense, lets merge this but I think that we can do a little better actually. I have been meaning to add an async client to aerial for a while now but this is great in the meantime

src/uagents/network.py
```diff
@@ -26,3 +37,26 @@ def get_faucet() -> FaucetApi:
...
+
+async def wait_for_tx_to_complete(
+    tx_hash: str,
+    timeout: Optional[timedelta] = None,
+    poll_period: Optional[timedelta] = None,
+) -> TxResponse:
+    if timeout is None:
+        timeout = timedelta(seconds=DEFAULT_QUERY_TIMEOUT_SECS)
+    if poll_period is None:
+        poll_period = timedelta(seconds=DEFAULT_QUERY_INTERVAL_SECS)
```
VijaySharma815, line 50:
```
timeout = timeout or timedelta(seconds=DEFAULT_QUERY_TIMEOUT_SECS)
poll_period = poll_period or timedelta(seconds=DEFAULT_QUERY_INTERVAL_SECS)
```

jrriehl: Good suggestion. This would work in almost all cases, except if someone wanted to set either value to 0 for some reason, so I'll probably leave it the way it is.

jrriehl:
> Yeah makes a lot of sense, lets merge this but I think that we can do a little better actually. I have been meaning to add an async client to aerial for a while now but this is great in the meantime

Agree 100%, was just thinking that we needed some intermediate solution while we work out the async client, since I'm not sure when that will happen.

PR #72, opened 2023-03-19
feat: protocol manifest generation

ejfitzgerald:
## Summary

First attempt at standardising manifest schema generation

Additionally I fixed a number of type issues that were in the project causing PyCharm to lose its mind a little


src/uagents/protocol.py
```diff
@@ -151,18 +151,68 @@ def _add_message_handler(
...
+                or request in self._signed_message_handlers
+            )
+
+            manifest["interactions"].append(
+                {
+                    "type": "query"
+                    if request in self._unsigned_message_handlers
+                    else "normal",
+                    "request": request,
+                    "responses": list(responses.keys()),
```
jrriehl, line 195: It would be kind of nice to have human readable model names here, too.

jrriehl (dismissed):

jrriehl (approved):

PR #77, opened 2023-03-24
feat: integrated name service contract

Alejandro-Morales:


MissingNO57 (dismissed):
LGTM

src/uagents/agent.py
```diff
@@ -232,12 +286,46 @@ def _schedule_registration(self):
...
+        query_msg = {
+            "permissions": {
+                "domain": f"{name}.agent",
+                "owner": {"address": {"address": str(self.wallet.address())}},
+            }
+        }
+        permission = self._service_contract.query(query_msg)["permissions"]
+        if permission == "admin":
+            return True
+        return False
```
jrriehl, line 327: It might make sense to have contract classes that inherit from `LedgerContract` for both the `almanac` and `name_service` contracts. Then we could add these methods to it like `name_service.is_owner(name: str, address: str)`.

jrriehl (commented):
Looks good, I'm just thinking in the interests of keeping the agent class as uncluttered as possible, it might be better to move all the contract stuff to `network.py`.

src/uagents/agent.py
```diff
@@ -343,9 +371,17 @@ def start_background_tasks(self):
 
         # start the contract registration update loop
         if self._endpoints is not None:
-            self._loop.create_task(
-                _run_interval(self._register, self._ctx, self._schedule_registration())
-            )
+            if (
+                not self._almanac_contract.is_registered(self.address)
+                or self._schedule_registration() < 3600
```
jrriehl, line 376: Should we maybe import 3600 as a constant from config.py?

jrriehl (dismissed):

examples/08-local-network-interaction/agent1.py
```diff
@@ -10,15 +10,22 @@ class Message(Model):
 
 
 bob = Agent(
-    name="bob",
+    name="bob123",
```
Autonomousweb3Dev, line 13: suggestion: "bob agent"

examples/08-local-network-interaction/agent1.py
```diff
@@ -10,15 +10,22 @@ class Message(Model):
...
     port=8001,
-    seed="bob secret phrase",
+    seed="bobs secret",
```
Autonomousweb3Dev, line 15: suggestion: bob agent secret pharse

src/uagents/resolver.py
```diff
@@ -2,43 +2,71 @@
...
+
+
 class Resolver(ABC):
     @abstractmethod
-    async def resolve(self, address: str) -> Optional[str]:
+    async def resolve(self, destination: str) -> Optional[str]:
         pass
 
 
 class AlmanacResolver(Resolver):
```
Dacksus, line 44: A comment towards the naming and potential future extensions. "AlmanacResolver" supposes this resolver is only using the almanac, when in fact it is either using the almanac or the name service. Wouldn't it be more consistent to rename this one to something like "GlobalResolver" that internally calls either the AlmanacResolver or the NameServiceResolver?

src/uagents/agent.py
```diff
@@ -207,50 +222,68 @@ async def _register(self, ctx: Context):
...
 
-        return sequence
+            return
+        transaction = prepare_and_broadcast_basic_transaction(
+            self._ledger, transaction, self.wallet
+        )
+        await wait_for_tx_to_complete(transaction.tx_hash)
+        self._logger.info("Registering name...complete")
+
+    def get_agent_address(self, name: str) -> str:
```
Dacksus, line 276: This function also exists in resolver.py. Are both needed or can the resolver one be called when needed?

Dacksus (changes requested):

src/uagents/resolver.py
jrriehl: It's a good point. I guess I thought that we could handle this automatically by detecting whether the resolver is trying to resolve an address or a name. But creating an explicit NameServiceResolver as a kind of wrapper for the AlmanacResolver might indeed a nicer way to do this.

src/uagents/network.py
```diff
@@ -24,7 +144,9 @@
 else:
     raise NotImplementedError
 
-_contract = LedgerContract(None, _ledger, CONTRACT_ALMANAC)
+
+_almanac_contract = AlmanacContract(None, _ledger, CONTRACT_ALMANAC)
+_service_contract = ServiceContract(None, _ledger, CONTRACT_SERVICE)
```
jrriehl, line 149: Probably more clear to call this `_name_service_contract`, `NameServiceContract`.

src/uagents/resolver.py
Alejandro-Morales: Yeah I think its a good idea, this is down now

Dacksus (approved):

PR #94, opened 2023-04-18
feat: bureau overwrites agent endpoints

jrriehl:
When creating a bureau of agents that should be reachable externally and are not using the mailbox, users should set the endpoint of the bureau and need not set endpoints for the agents. Example:
```python
from uagents import Agent, Bureau

agent = Agent(name="agent")

port = 8000
bureau = Bureau(endpoint=f"http://127.0.0.1:{port}/submit", port=port)
bureau.add(agent)
bureau.run()
```

### DIFF
--- a/src/uagents/agent.py
+++ b/src/uagents/agent.py
@@ -429,7 +429,7 @@ class Bureau:
     def __init__(
         self,
         port: Optional[int] = None,
-        endpoint: Optional[Union[List[str], Dict[str, dict]]] = None,
+        endpoint: Optional[Union[str, List[str], Dict[str, dict]]] = None,
     ):
         self._loop = asyncio.get_event_loop_policy().get_event_loop()
         self._agents = []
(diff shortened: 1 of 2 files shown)


src/uagents/config.py
```diff
@@ -38,6 +38,8 @@ def parse_endpoint_config(
         ]
     elif isinstance(endpoint, list):
         endpoints = [{"url": val, "weight": 1} for val in endpoint]
+    elif isinstance(endpoint, str):
```
jrriehl, line 41: This seems like an obvious omission in the API for specifying endpoints.

src/uagents/agent.py
```diff
@@ -423,9 +426,14 @@ async def _process_message_queue(self):
 
 
 class Bureau:
-    def __init__(self, port: Optional[int] = None):
+    def __init__(
+        self,
+        port: Optional[int] = None,
+        endpoint: Optional[Union[List[str], Dict[str, dict]]] = None,
```
ejfitzgerald, line 432:
```suggestion
        endpoint: Optional[Union[str, List[str], Dict[str, dict]]] = None,
```
To match your other change

ejfitzgerald (approved):

PR #96, opened 2023-05-04
deps: update cosmpy to 0.7.0

jrriehl:


pyproject.toml
```diff
@@ -7,15 +7,15 @@ license = "Apache 2.0"
 readme = "README.md"
 
 [tool.poetry.dependencies]
-python = "^3.8"
+python = ">=3.8,<3.11"
```
jrriehl, line 10: This was needed for compatibility with cosmpy 0.7.0, but we should try to remove this restriction in both soon.

Alejandro-Morales (approved):

PR #99, opened 2023-05-23
feat: protocol-aware envelopes

jrriehl:
When a message sent is part of a protocol used by the agent, it can be useful to see this information in the envelope. Then the protocol manifest can be looked up from the digest, which gives much more context for a specific message.

In this PR:
- Envelope `protocol` field is now called `schema_digest` (keeping `protocol` as an alias)
- Adds optional `protocol_digest` field for the digest of the protocol to which the message belongs (if available)
- `Context.send` now tries to look up the protocol for a message before sending it
- Adds optional `nonce` field to allow for prevention of replay attacks
- This should be fully backwards compatible


Alejandro-Morales (dismissed):
Looks good 👍

PR #100, opened 2023-05-23
feat: add protocol query in contrib module

jrriehl:
Adds a built-in protocol for responding to protocol queries, that is, requests to retrieve the manifest of a particular protocol, or if none is specified, all protocols included in the agent are included in the response.


Archento (approved):
LGTM 👍🏻

PR #104, opened 2023-06-14
feat: use session id correctly

jrriehl:
Use session ID of incoming messages whenever sending from a message handler.

Adds session to context so it will be available in message handlers, too.

### DIFF
--- a/src/uagents/context.py
+++ b/src/uagents/context.py
@@ -61,7 +61,7 @@ class Context:
         self._resolver = resolve
         self._identity = identity
         self._queries = queries
-        self._session = session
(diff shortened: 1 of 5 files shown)


src/uagents/context.py
```diff
@@ -60,6 +61,7 @@ def __init__(
         self._resolver = resolve
         self._identity = identity
         self._queries = queries
+        self._session = session
```
ejfitzgerald, line 64: I think that 'ctx.session' should always return a UUID. So I would maybe generate a new one here is 'session is None'

ejfitzgerald (approved):

PR #107, opened 2023-06-20
feat: add context.send_raw function

jrriehl:
Adds a `send_raw` function to send Json-encoded messages along with a schema digest for when the Pydantic Model is not available. This small refactor should be completely backwards compatible.


qati (approved):
LGTM

ejfitzgerald (approved):

PR #109, opened 2023-06-22
feat: publish protocol manifest to agentverse

jrriehl:
Adds option to publish protocol manifest when including it in an agent.

Also updates the config so that the agentverse URL can be specified without turning on the mailbox client.


Autonomousweb3Dev (approved):

NindoK (dismissed):

jrriehl:
> LGTM overall. I am a bit confused to the use of both agentverse and mailbox parameters to Agent() . If I understood correctly, agentverse is supposed to replace mailbox, but we are keeping mailbox for backward compatibility. Assuming the above correct, it would be better in my opinion to make them mutually exclusive and to print a warning message stating that mailbox is deprecated in favour for agentverse when mailbox is used

Thanks for the review. I think we are basically in agreement. I did make them mutually exclusive by raising an exception when both `mailbox` and `agentverse` are passed to `Agent()`: https://github.com/fetchai/uAgents/pull/109/files#diff-5a21004e6c4cd358b9dea71e3494d5b78b052f89b1d7f223fe4fc69942bb4184R92-R95. 

But I will update the error message and add a deprecation warning when only mailbox is used.

lrahmani (approved):
LGTM, thanks

PR #111, opened 2023-07-03
feat: add descriptions without modifying model digest

zmezei:


src/uagents/models.py
```diff
@@ -1,20 +1,61 @@
...
+    ):
+        for field_name, field in model.__fields__.items():
+            if field.field_info and field.field_info.description:
+                orig_descriptions[field_name] = field.field_info.description
+                field.field_info.description = None
+            elif issubclass(field.type_, Model):
+                orig_descriptions[field_name] = {}
+                Model._remove_descriptions(field.type_, orig_descriptions[field_name])
+
+    @staticmethod
```
dominic22, line 23: From my point of view these methods shouldn't be static  most of the times the model is even passed in why not use `self` then?

src/uagents/protocol.py
```diff
@@ -176,7 +176,7 @@ def manifest(self) -> Dict[str, Any]:
 
         for schema_digest, model in all_models.items():
             manifest["models"].append(
-                {"digest": schema_digest, "schema": model.schema()}
+                {"digest": schema_digest, "schema": model.schema_no_descriptions}
```
lrahmani, line 179: Is it guaranteed that this value won't be None?

src/uagents/protocol.py
```diff
@@ -199,6 +199,12 @@ def manifest(self) -> Dict[str, Any]:
         encoded = json.dumps(manifest, indent=None, sort_keys=True).encode("utf8")
         metadata["digest"] = f"proto:{hashlib.sha256(encoded).digest().hex()}"
```
lrahmani, line 200: digest is also computed by the `compute_digest` static method.
It takes as input the manifest which will contain the descriptions hence producing a different digest.

lrahmani (changes requested):
I believe this feature needs to be rethought as it can produce different digests for the same model. Also removing model fields may not be the best approach to skip them in digest computation. One can simply delete the fields from the manifest dict.

src/uagents/models.py
zmezei: The model param passed to these newly created methods is actually not an instance of a Model subclass but rather a Model subclass type (so a class object) that's why I couldn't implement these methods as instance methods (so using self.). However you are right in the sense that these method could be class methods using cls.

src/uagents/protocol.py
zmezei: This class attribute is set when generating model digest and we calculate the digest for every model beforehand but for safety I can modify this line to:
`model.schema_no_descriptions if model.schema_no_descriptions is not None else model.schema()`
Does it make sense?

src/uagents/protocol.py
zmezei: I modified this method accordingly and also created test for it in test_fields_descr.py, please check it.

src/uagents/models.py
zmezei: In the refactored version I created class methods instead of static methods where possible, please check it.

src/uagents/protocol.py
zmezei: In the refactored version I call the schema_no_descriptions(), that method also uses a class attribute as a cache and that class attribute could possibly be None, but in the schema_no_descriptions() method I guarantee to set the value of that class variable now if that is None.

PR #114, opened 2023-07-13
feat: new almanac and name service registration flow

Alejandro-Morales:


src/uagents/agent.py
```diff
@@ -258,8 +258,14 @@ async def register_name(self):
...
+                f"Domain {domain} is not public, please select a public domain"
+            )
+            return
+
         transaction = self._service_contract.get_registration_tx(
-            self.name, str(self.wallet.address()), self.address
+            self.name, str(self.wallet.address()), self.address, domain
         )
 
         if transaction is None:
```
jrriehl, line 271: Could we also check this beforehand using `is_name_available(name, domain)`?

jrriehl:
I updated this PR with a few things:

1. Almanac registration is now performed before startup tasks. This means that actions performed `on_event("startup")` can now assume that registration has already taken place. This will help to prevent errors when returning messages are undeliverable.

2. Name service registration example now uses a simple CosmPy owner wallet to illustrate the fact that domain name owners need not be and will generally not be the agents themselves.

3. All of the contract interactions are moved to `uagents.network`.

examples/13-agent-name-service/agent1.py
```diff
@@ -0,0 +1,44 @@
...
+name_service_contract = get_name_service_contract()
+DOMAIN = "agent"
+
+for wallet in [my_wallet, bob.wallet]:
+    fund_agent_if_low(wallet.address())
+
+
+@bob.on_event("startup")
+async def register_agent_name(ctx: Context):
+    await name_service_contract.register(
```
lrahmani, line 33: I see that the name service registration is not handled implicitly by the library unlike almanac registration.
Not sure if it is better to have it as part of the agent handlers or to put it outside similar to `agent.run()`

lrahmani (approved):
LGTM, just a minor inline comment

examples/13-agent-name-service/agent1.py
jrriehl: Right, we originally had it inside the agent similar to the Almanac registration, but decided to move it out because it seemed to make more sense when being sent from an external wallet.

PR #116, opened 2023-07-27
feat: add broadcast to agents via protocol

Archento:


src/uagents/context.py
```diff
@@ -97,22 +99,61 @@ def get_message_protocol(self, message_schema_digest) -> Optional[str]:
...
+
     async def send(
         self,
         destination: str,
         message: Model,
         timeout: Optional[int] = DEFAULT_ENVELOPE_TIMEOUT_SECONDS,
     ):
         schema_digest = Model.build_schema_digest(message)
-        await self.send_raw(
+        await self._send_raw(
```
jrriehl, line 125: We probably don't want to make this private because some people are using `ctx.send_raw()` in message handlers.

src/uagents/config.py
```diff
@@ -27,6 +27,7 @@ class AgentNetwork(Enum):
 AGENT_NETWORK = AgentNetwork.FETCHAI_TESTNET
 
 AGENTVERSE_URL = "https://agentverse.ai"
+ALMANAC_API_URL = "https://staging.agentverse.ai/v1/almanac/"
```
jrriehl, line 30: Can we just use `AGENTVERSE_URL + "/v1/almanac"` instead?

The intention was that if people want to use the staging environment, they can create their agents with:
```python
agent = Agent(agentverse="https://staging.agentverse.ai")
```

jrriehl (commented):
A couple of minor comments, but otherwise looks great!

src/uagents/config.py
Dacksus: Makes sense!

Is the almanac API already reachable on the production environment, or do we need to relay all agents to staging for the time being?

jrriehl: Yes, it should be fine on production.

src/uagents/context.py
```diff
@@ -97,6 +99,22 @@ def get_message_protocol(self, message_schema_digest) -> Optional[str]:
                     return protocol_digest
         return None
 
+    def get_agents_by_protocol(self, protocol_digest: str) -> list:
```
ejfitzgerald, line 102:
```suggestion
    def get_agents_by_protocol(self, protocol_digest: str, limit: Optional[int] = None) -> List[str]:
```

The return type should be a bit more specific, also there needs to be a `limit` in the API for the maximum number of entries that are returned.

src/uagents/context.py
```diff
@@ -97,6 +99,22 @@ def get_message_protocol(self, message_schema_digest) -> Optional[str]:
...
+        if not isinstance(protocol_digest, str) or not protocol_digest.startswith(
+            "proto:"
+        ):
+            self.logger.error(f"Invalid protocol digest: {protocol_digest}")
+            return []
```
ejfitzgerald, line 107: This doesn't really feel like a it should be a silent error to me. If the user is not providing the right values this seems like this should `raise` some sort of `ValueError` - that would be clearer I think

src/uagents/context.py
```diff
@@ -112,6 +130,29 @@ async def send(
             timeout=timeout,
         )
 
+    async def broadcast(
+        self,
+        destination: str,
+        message: Model,
+        timeout: Optional[int] = DEFAULT_ENVELOPE_TIMEOUT_SECONDS,
+    ):
```
ejfitzgerald, line 138:
```suggestion
    async def experimental_broadcast(
        self,
        destination_protocol: str,
        message: Model,
        limit: Optional[int] = DEFAULT_SEARCH_LIMIT,
        timeout: Optional[int] = DEFAULT_ENVELOPE_TIMEOUT_SECONDS,
    ):
```

A few things:

* Parameter naming to make it more explicit.
* Related to the above we need to have a limit in the function
* Lastly, I think that we need some way to mark new functions as experimental or to mark that they are not part of the stable API (yet). I am not sure if this broadcast pattern is something that longer term that we will need, however, I think if we mark it as experimental then we can hedge our bets a little. @jrriehl / @lrahmani do you agree?

src/uagents/context.py
```diff
@@ -112,6 +130,29 @@ async def send(
...
+            try:
+                await self.send_raw(
+                    address,
+                    message.json(),
+                    schema_digest,
+                    message_type=type(message),
+                    timeout=timeout,
+                )
+            except asyncio.exceptions.CancelledError:
+                self.logger.error(f"Error sending message to {address}")
```
ejfitzgerald, line 154: Potentially consider gathering all these awaitables together:

https://docs.python.org/3/library/asyncio-task.html#asyncio.gather

Might make the error handling a little more complicated. Let me know what you think.

Archento: Gather is a good idea I think as it should reduce complexity in the task queue. As for the error handling, this is a topic that should be tackled on a broader level first I think. James confirmed that this is something that was planned for the current/upcoming sprint, after that we can refine exception handling a layer above, e.g. in this broadcast method.

src/uagents/context.py
jrriehl: This could be addressed by retaining `experimental_broadcast` for some deprecation period as a simple redirect to `broadcast`.

PR #126, opened 2023-08-14
feat: restructure repository and add integrations

jrriehl:
- Move Python library into `python` folder
- Update README files
- Add `integrations` folder with initial `fetch-holiday` integration


.github/workflows/ci-tests.yml
```diff
@@ -15,6 +15,7 @@ jobs:
     defaults:
```
lrahmani, line 15: You would want to add `path` to `on: push:` to run the workflow only when files in `python` are updated

jrriehl: Do you think we need on `push` if we also run this on `pull_request`?

GataKamsky:
Looks good from my perspective. @ejfitzgerald you good with this?

GataKamsky (approved):
Good from my perspective.

.github/workflows/ci-tests.yml
```diff
@@ -1,20 +1,20 @@
 name: CI
 
 on:
-  push:
-    branches:
-      - main
-
```
ejfitzgerald, line 7: Would still keep this. Worth running full checks on each commit to `main`

ejfitzgerald (commented):
Looks good one minor CI comment

ejfitzgerald (approved):

PR #133, opened 2023-08-24
chore: handle exceptions in message handlers

jrriehl:
Generally tried to ensure we catch exceptions wherever a task is awaited or created.

### DIFF
--- a/python/src/uagents/agent.py
+++ b/python/src/uagents/agent.py
@@ -2,6 +2,7 @@ import asyncio
 import functools
 from typing import Dict, List, Optional, Set, Union, Type, Tuple, Any, Coroutine
 import uuid
+from pydantic import ValidationError
 import requests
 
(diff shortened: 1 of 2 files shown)


Archento (dismissed):
The added try-catch blocks all look good to me.

I'm just wondering if catching those specific errors at the agent level and forwarding them to the logger is enough or if we should also try-catch at a deeper level like context.py:L245-254.
Just an idea, otherwise I think it's a good start :)

jrriehl:
> The added try-catch blocks all look good to me.
> 
> I'm just wondering if catching those specific errors at the agent level and forwarding them to the logger is enough or if we should also try-catch at a deeper level like context.py:L245-254. Just an idea, otherwise I think it's a good start :)

Good point, I added some exception handling here as well.

Archento (dismissed):
Yes, that's what I had in mind! Good stuff :)

PR #141, opened 2023-09-04
feat: add release automation

jrriehl:
Add release automation script and workflow (copied over from https://github.com/fetchai/python-template).


.github/workflows/release.yml
```diff
@@ -0,0 +1,34 @@
+name: Check and release uAgents
+# runs release script on updated main branch
+# script checks for current package version and the latest tag on github
+# if package version is newer it builds, publishes to PyPI, tags and makes a release page on github
+# secrets PYPI_USERNAME and PYPI_PASSWORD are required!
```
5A11, line 5: Are the secrets configured?

.github/workflows/release.yml
```diff
@@ -0,0 +1,34 @@
...
+      run:
+        shell: bash
+        working-directory: ./python
+    if: github.event.pull_request.merged == true && startsWith(github.event.pull_request.head.ref, 'release')
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-python@v4
+      with:
+          python-version: '3.9'
```
5A11, line 23: change to latest (3.11?)

5A11 (commented):
I suggest you replace these files with those in https://github.com/fetchai/python-template. This version doesn't need you to maintain a History.md in the repo and does some additional stuff (auto-generates gh release notes, marks it as latest release, etc).

.github/workflows/release.yml
```diff
@@ -0,0 +1,34 @@
...
+    - name: Run release script
+      env:
+        PYPI_USERNAME: __token__
+        PYPI_PASSWORD: ${{secrets.FETCHBOT_PYPI_TOKEN}}
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}  # required to make a release page
```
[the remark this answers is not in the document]
jrriehl: Yes

5A11:
> > I suggest replacing these files with those in https://github.com/fetchai/python-template. You would only need the `do-release` and the `release.yml` workflow files as this version doesn't need you to maintain a History.md in the repo and does some additional stuff (auto-generates gh release notes, marks it as latest release, etc).
> 
> Good suggestion. How does it do the release notes? Are they taken straight from the PR commit messages?

It should be the same as when you auto-fill the release notes on GitHub (taken from PR titles/commit message). 

A fair warning, I haven't tested the release part on the python-template repo with an actual release. Though the changes from the cosmpy version which we know works are minor and mostly clean up so I wouldn't expect a high likelihood of failure. In any case, wanted to give you a heads up in case you want to test it with a minor release.

5A11 (approved):
LGTM. Just note the instruction at the top of the workflow file with the branch names and ensure you have the secretes set up.

PR #160, opened 2023-09-17
openai whisper model integration

abhi141:
openai whisper model integration with microAgent, user can use this to transcript audio recording easily.


jrriehl (approved):
Nice work, LGTM!

PR #163, opened 2023-09-18
fix: make http headers case insensitive

jrriehl:
Uses the `CaseInsensitiveDict` from `requests.structures`.


MissingNO57 (approved):
LGTM

PR #165, opened 2023-09-18
add broadcast by protocol example

Archento:


python/examples/14-broadcast/main.py
```diff
@@ -0,0 +1,58 @@
...
+class Request(Model):
+    pass
+
+
+class Response(Model):
+    text: str
+
+
+# define protocol
+proto = Protocol(name="proto", version="1.0")
```
jrriehl, line 25: Suggest giving these all more descriptive names like `BroadcastExampleRequest` to reduce the chance that this example will send messages to other agents with a generic Request/Response protocol.

Archento: You are absolutely right. Will change those names.

PR #166, opened 2023-09-19
fix: improve request header parsing

jrriehl:
- Fixes issue where connecting from browser results in an uncaught 500 error
- Returns status OK when connecting from browser
- More informative error messages


Archento (approved):
Looks good! I like the more precise responses.
Also: It works as expected :)

PR #169, opened 2023-09-22
fix: use weighted random sampling without replacement

jrriehl:
Endpoints returned from the Almanac contract were getting chosen multiple times because `random.choices` replaces values before each weighted random choice. Such a sampling method is not included in the Python standard library, but I found a well-sourced implementation in pure Python here: https://maxhalford.github.io/blog/weighted-sampling-without-replacement/.


Archento (dismissed):
Good find!

PR #173, opened 2023-09-27
feat: update agent addresses

Alejandro-Morales:
Agents have a new `test `attribute that will default True when initialized. This indicates if agent will be registered on testnet or mainnet contracts.

When agents want to communicate with other agents, they can use `agent.identifier` that will contain a network prefix + agent.address:

Addresses with prefix:

- `test-agent://` will resolve via testnet contracts
- `agent:// `will resolve via mainnet contracts
- agent addresses without any prefix will resolve to testnet to preserve backwards compatibility


python/examples/13-agent-name-service/agent1.py
```diff
@@ -19,19 +23,25 @@ class Message(Model):
...
+fund_agent_if_low(bob)
+
 my_wallet = LocalWallet.from_unsafe_seed("registration test wallet")
 name_service_contract = get_name_service_contract()
 DOMAIN = "agent"
 
-for wallet in [my_wallet, bob.wallet]:
-    fund_agent_if_low(wallet.address())
+faucet = get_faucet()
+AGENT_BALANCE = bob._ledger.query_bank_balance(my_wallet)
```
jrriehl, line 33: How about adding a `balance()` method directly to the agent?

python/src/uagents/setup.py
```diff
@@ -20,16 +21,26 @@ def fund_agent_if_low(wallet_address: str):
     Returns:
         None
     """
-    ledger = get_ledger()
+    ledger = get_ledger(agent._test)
```
jrriehl, line 24: Let's add a `ledger` property to the agent so that we can just use `agent.ledger`.

python/src/uagents/setup.py
```diff
@@ -20,16 +21,26 @@ def fund_agent_if_low(wallet_address: str):
...
     faucet = get_faucet()
 
-    agent_balance = ledger.query_bank_balance(Address(wallet_address))
+    agent_balance = ledger.query_bank_balance(Address(agent.wallet.address()))
```
jrriehl, line 27: Maybe it is actually useful to add a `balance` property to the agent too.

python/src/uagents/config.py
```diff
@@ -1,22 +1,24 @@
 import logging
 import sys
-from enum import Enum
+
+# from enum import Enum
```
jrriehl, line 4:
```suggestion
```

python/examples/13-agent-name-service/agent1.py
Alejandro-Morales: Ok If we can assume that we could import testnet ledger from network. And yeah incorrect variable name, should have been WALLET_BALANCE

python/src/uagents/agent.py
```diff
@@ -467,7 +505,7 @@ async def register(self):
             or list(self.protocols.keys())
             != self._almanac_contract.get_protocols(self.address)
         ):
-            agent_balance = self._ledger.query_bank_balance(
+            agent_balance = self.ledger.query_bank_balance(
```
jrriehl, line 508: Can now be updated to `self.balance` right?

python/examples/06-send-tokens/main.py
```diff
@@ -2,6 +2,8 @@
 from uagents.network import wait_for_tx_to_complete
 from uagents.setup import fund_agent_if_low
 
+# pylint: disable=protected-access
```
jrriehl, line 5: Should be safe to remove this now I think
```suggestion
```

python/src/uagents/agent.py
Alejandro-Morales: Thats right

python/src/uagents/context.py
```diff
@@ -374,6 +386,10 @@ async def send_raw(
         Returns:
             MsgStatus: The delivery status of the message.
         """
+
+        # Destination without ledger prefix
+        raw_destination = destination[-65:]
```
jrriehl, line 391: Things get a bit complicated in here because destination can also be a name at this point. I guess it's what we're calling an `identifier` now, and will need a way to parse this. The resolver should handle most of it, but we might need a simple function to extract the agent address from an identifier for things like checking if it is registered with the local dispatcher.

python/src/uagents/context.py
```diff
@@ -391,7 +407,7 @@ async def send_raw(
                     return MsgStatus(
                         status=DeliveryStatus.FAILED,
                         detail="Invalid reply",
-                        destination=destination,
+                        destination=raw_destination,
```
jrriehl, line 410: I think we can probably leave these as `destination` / `identifier`.

python/src/uagents/context.py
```diff
@@ -403,34 +419,34 @@ async def send_raw(
...
                     status=DeliveryStatus.FAILED,
                     detail="Invalid interval message",
-                    destination=destination,
+                    destination=raw_destination,
                     endpoint="",
                 )
 
         # Handle local dispatch of messages
-        if dispatcher.contains(destination):
+        if dispatcher.contains(raw_destination):
```
jrriehl, line 427: Here we need the `address`.

python/examples/13-agent-name-service/agent1.py
```diff
@@ -1,9 +1,11 @@
 from cosmpy.aerial.wallet import LocalWallet
 
-from uagents.network import get_ledger, get_name_service_contract
+from uagents.network import get_faucet, get_name_service_contract, _testnet_ledger
```
jrriehl, line 3: Can we import `get_ledger` here instead?

python/src/uagents/context.py
Alejandro-Morales: @jrriehl I updated the code with a proposal for this, let me know what you think

python/src/uagents/context.py
```diff
@@ -45,6 +48,29 @@
...
+    prefixes = [TESTNET_PREFIX, MAINNET_PREFIX, ""]
+    expected_length = 65
+
+    for prefix in prefixes:
+        if destination.startswith(prefix) and len(destination) == expected_length + len(
+            prefix
+        ):
+            return destination[len(prefix) :]
+
+    return None
```
jrriehl, line 71: I guess the best approach for this kind of function that can have multiple possible inputs is to write tests for each of the possible inputs. Here's what I think are the possibilities:
```python
agent1blahblahblah -> agent1blahblahblah
some-prefix://agent1blahblahblah -> agent1blahblahblah
some-prefix://some_agent_name/agent1blahblahblah -> agent1blahblahblah
```
This functions looks like it would work on the first two, but probably not on the third. I realize we haven't implemented the resolver logic for agent names in the identifier yet, so just looking a bit ahead here.

For these 3 cases, I think something as simple as:
```python
return destination.split("://")[-1].split("/")[-1]
```
could work, but needs to be tested.

Alejandro-Morales: Ok I guess I didnt consider the other scenarios. Ill try this out

python/src/uagents/setup.py
jrriehl: We shouldn't really use private class members outside the class. That's kind of the point of them and the reason that pylint check is useful. If we really need access to it, then we can make a property or make it public, but in this case, since there is no such thing as a mainnet faucet, then it only makes sense to use the testnet faucet here:
```
ledger = get_ledger(test=True)
```
But I think we should also log a warning if an agent who's ledger is set to mainnet tries to get funds, to make it clear that this function is only for adding funds on the testnet.

One more thing is that this is a pretty big change that will break a lot of people's code, so I think we should make it backwards compatible. If the argument passed is an address, then we should simply add testnet funds to that address.

python/examples/01-first-agent/main.py
```diff
@@ -5,7 +5,7 @@
 
 @agent.on_event("startup")
 async def introduce_agent(ctx: Context):
-    ctx.logger.info(f"Hello, I'm agent {ctx.name} and my address is {ctx.address}.")
+    ctx.logger.info(f"Hello, I'm agent {ctx.name} and my address is {ctx.identifier}.")
```
jrriehl, line 8: I guess we should be consistent with what we call this in the logs, so either print the address here (I think that's slightly better) or call this the identity in the log statement.

python/examples/02-interval-task/main.py
```diff
@@ -5,7 +5,7 @@
 
 @agent.on_event("startup")
 async def introduce_agent(ctx: Context):
-    ctx.logger.info(f"Hello, I'm agent {ctx.name} and my address is {ctx.address}.")
+    ctx.logger.info(f"Hello, I'm agent {ctx.name} and my address is {ctx.identifier}.")
```
jrriehl, line 8: Same comment as above.

python/examples/05-send-msg/main.py
```diff
@@ -12,7 +12,7 @@ class Message(Model):
 @alice.on_interval(period=2.0)
 async def send_message(ctx: Context):
     msg = f"Hello there {bob.name} my name is {alice.name}."
-    await ctx.send(bob.address, Message(text=msg))
+    await ctx.send(bob.identifier, Message(text=msg))
```
jrriehl, line 15: I kind of agree that this makes more sense in general, but for sending messages within a bureau, the identifier is not actually useful, so I would probably leave this as `bob.address`. It's tricky because we want to introduce the concept of identifier in a way that doesn't seem like we are changing something fundamental, which it really isn't. So I'd suggest using identifier only in the examples that actually use the Almanac and name service contracts.

Alejandro-Morales: I agree

python/src/uagents/resolver.py
```diff
@@ -160,7 +198,9 @@ def __init__(self, max_endpoints: Optional[int] = None):
         self._max_endpoints = max_endpoints or DEFAULT_MAX_ENDPOINTS
         self._almanac_resolver = AlmanacResolver(max_endpoints=self._max_endpoints)
 
-    async def resolve(self, destination: str) -> Tuple[Optional[str], List[str]]:
+    async def resolve(
+        self, destination: str, test: bool
```
jrriehl, line 202: Nitpicking a bit, but ideally `Resolver.resolve(...)` should take the same arguments for each resolver, as defined in the abstract `Resolver` class:
```
async def resolve(self, destination: str) -> Tuple[Optional[str], List[str]]:
```
And now that we have the concept of identifier, I think we could call it that in the functions:
```
async def resolve(self, identifier: str) -> Tuple[Optional[str], List[str]]:
```
Now it is clear that the `identifier` should have the all the information needed to resolve the endpoint, so it is the job of this `resolve` function to determine whether to go to the testnet or mainnet contracts.  Does that make sense?

python/src/uagents/resolver.py
```diff
@@ -4,11 +4,38 @@
...
-from uagents.config import DEFAULT_MAX_ENDPOINTS
+from uagents.config import (
+    DEFAULT_MAX_ENDPOINTS,
+    TESTNET_PREFIX,
+    MAINNET_PREFIX,
+    AGENT_PREFIX,
+)
 from uagents.network import get_almanac_contract, get_name_service_contract
 
+# pylint: disable=arguments-differ
```
jrriehl, line 15: I think this will not be needed if the update from my comment below is implemented. In general these pylint rules are there for good reasons and we should have a better reason to ignore them, which is sometimes the case, but I'm not convinced it is here.

python/src/uagents/resolver.py
Alejandro-Morales: Yes, I was actually making that fix already 👍

jrriehl (approved):
Looks good!

PR #176, opened 2023-10-04
feat: add default error handler

jrriehl:
- Renames misnamed `_handle_error` function to `_send_error_message`
- Adds a default error handler that can easily be overwritten if desired

A possible bonus is that rather than an empty protocol being default, the default protocol will now have something in it.

### DIFF
--- a/python/src/uagents/agent.py
+++ b/python/src/uagents/agent.py
@@ -74,7 +74,7 @@ async def _delay(coroutine: Coroutine, delay_seconds: float):
     await coroutine
 
 
-async def _handle_error(ctx: Context, destination: str, msg: ErrorMessage):
+async def _send_error_message(ctx: Context, destination: str, msg: ErrorMessage):
     """
     Send an error message to the specified destination.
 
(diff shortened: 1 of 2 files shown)


python/src/uagents/agent.py
```diff
@@ -255,6 +255,11 @@ def __init__(
                 self._port, self._loop, self._queries, logger=self._logger
             )
 
+        # define default error message handler
+        @self.on_message(ErrorMessage)
+        async def _handle_error_message(ctx: Context, sender: str, msg: ErrorMessage):
+            ctx.logger.warning(f"Received error message from {sender}: {msg.error}")
```
Archento, line 261: Why not use `logger.error()`? It could be misleading when the text refers to an error message and the agent only logs a warning.

Archento (approved):
Apart from my one question this works perfectly as intended. Overwrite/rewrite of the internal handler is also easily possible. 
Very happy with the quick solution :)

python/src/uagents/agent.py
jrriehl: Yes, agreed. I was thinking that errors coming back from another agent might not rise to the level of an internal agent exception, but it's more likely that an error like this does indicate a problem in the agent who's receiving the error message. Updated to `logger.exception`.

Archento (approved):

PR #180, opened 2023-10-08
docs: fixed broken examples link

bgoober:
The main branch's examples link did not include the parent '/python' folder. I have replaced it with the correct '../python/examples' URL.


jrriehl (approved):
Good catch, thanks!

PR #186, opened 2023-10-16
feat: add stateful communication in experimental module

Archento:
Current state of the implementation:

- add `Dialogue` class to experimental module
- add example 15 that makes use of the new dialogue class
- modify context and agent files to accommodate the new functionality


Archento:
Most of the logic is inside `python/src/uagents/experimental/dialogues/__init__.py` but GitHub decides to collapse the file by default for some reason in the 'Files changed' view so please keep that in mind when reviewing.

python/src/uagents/experimental/dialogues/__init__.py
```diff
@@ -0,0 +1,467 @@
...
+        if session in cache.keys():
+            cache.pop(session)
+        self._storage.set(self.name, cache)
+
+    def _update_transition_model(self, edge: Edge, model: Type[Model]) -> None:
+        """Update the message model for a transition."""
+        self._mapping[edge.name] = Model.build_schema_digest(model)
+        self._edges[self._edges.index(edge)].model = model
+
+    def on_state_transition(self, edge_name: str, model: Type[Model]):
```
jrriehl, line 417: I think there is a case for simply overriding `on_message` here and switching the order of `edge` and `model`.

If you do that, then the development pattern for adding a message handler is very similar to before, where the concept of `state_transition` is introduced as a required argument whenever you use the `Dialogue` class.

This also avoids having to disable `on_message` somehow.

But presumably you've already considered this, so maybe I am missing some disadvantage of doing this.

python/src/uagents/experimental/dialogues/__init__.py
```diff
@@ -0,0 +1,467 @@
...
+    - A storage will contain all the dialogues that took place, which may be
+      automatically deleted after a certain amount of time.
+    - is meant to simplify the handling of individual messages
+    """
+
+    def __init__(
+        self,
+        name: str,  # mandatory, due to storage naming
+        version: Optional[str] = None,
+        nodes: list[Node] | None = None,
```
jrriehl, line 89: To maintain support for Python 3.8 and 3.9, we need to use the older notion, importing from the `typing` library.
```suggestion
        nodes: List[Node] | None = None,
```

python/src/uagents/agent.py
```diff
@@ -880,6 +882,46 @@ async def _process_message_queue(self):
...
+                                self._ctx.logger.debug("dialogue picked up")
+
+                            context.dialogue = protocol.get_session(session)
+                            protocol.add_message(
+                                session_id=session,
+                                message_type=self._models[schema_digest].__name__,
+                                sender=sender,
+                                receiver=self.address,
+                                content=message,
+                            )
```
jrriehl, line 924: Feels like this block should go in a dedicated function somewhere, maybe in `Dialogue`?

python/examples/15-chit-chat-dialogue/dialogues/simple_chitchat.py
```diff
@@ -0,0 +1,101 @@
...
+# Node definition for the dialogue states
+waiting_state = Node(
+    name="Waiting",
+    description=(
+        "This is the default state of the dialogue. Every session starts in "
+        "this state and is automatically updated once a transition to start or "
+        "continue a dialogue is triggered."
+    ),
+    starter=True,
+)  # currently not used as states are measured by the edges
```
jrriehl, line 13: Would it make any sense to define nodes and edges as types that inherit from `Node` and `Edge`?
```python
class WaitingState(Node):
    name="Waiting"
    description=(
        "This is the default state of the dialogue. Every session starts in "
        "this state and is automatically updated once a transition to start or "
        "continue a dialogue is triggered."
    )
    starter=True
)  # currently not used as states are measured by the edges

class InitSession(Edge):
     name="Initiate Session"
     description="This is the transition to start the dialogue"
     parent: None
     child: InitState
```
This may not be standard practice, because I'm not sure whether there would be a use for instantiating one of these, but architecturally it seems like it could be sensible. This way you could write something like:
```python
@dialogue.on_message(model=InitiateChitChatDialogue, state_transition=InitSession)
```
Just an idea though, I haven't thought this all the way through.

jrriehl (commented):
Really nice implementation! Just a few comments / thoughts for discussion.

python/src/uagents/agent.py
Archento: Good point, will be transferred to `Dialogue`.

python/examples/15-chit-chat-dialogue/dialogues/simple_chitchat.py
Archento: To keep things as familiar as possible we absolutely agree with this approach. Will be changed :)

python/src/uagents/experimental/dialogues/__init__.py
Archento: OK, changed

python/src/uagents/experimental/dialogues/__init__.py
```diff
@@ -0,0 +1,467 @@
...
+        timeout: int = DEFAULT_SESSION_TIMEOUT,
+        max_nr_of_messages: int = 100,
+    ) -> None:
+        self._name = name
+
+        self._nodes = nodes or []
+        self._edges = edges or []
+        self._graph: dict[str, list[str]] = self._build_graph()  # by nodes
+        self._rules: dict[str, list[str]] = self._build_rules()  # by edges
+        self._mapping: dict[str, str] = {
```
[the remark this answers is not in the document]
Archento: changed

python/src/uagents/agent.py
Archento: After a bit of consideration I think it would be good to keep this code in `agent.py` because about half of the lines is just debugging messages. The rest accesses the context or agent and moving it to the `Dialogue` class would add unnecessary complexity.
We can remove the debug messages later or add additional verbosity throughout the framework but in principle I agree that it otherwise creates a logging imbalance.

Dacksus:
needed to undraft:

- [ ] refine example description / documentation
- [ ] debug dialogue.update_state not working anymore

python/examples/17-stateful-communication/main.py
```diff
@@ -0,0 +1,218 @@
+"""Chit chat dialogue example"""
+import json
+from uuid import uuid4
+
+from uagents import Agent, Bureau, Context, Model
+from uagents.setup import fund_agent_if_low
```
jrriehl, line 6:
```suggestion
```

python/examples/17-stateful-communication/main.py
```diff
@@ -0,0 +1,218 @@
...
+
+from dialogues.chitchat import ChitChatDialogue
+
+agent1 = Agent(name="agent1", seed="9876543210000000000")
+agent1._logger.setLevel("DEBUG")  # pylint: disable=protected-access
+fund_agent_if_low(agent1.wallet.address())
+
+agent2 = Agent(name="agent2", seed="9876543210000000001")
+agent2._logger.setLevel("DEBUG")  # pylint: disable=protected-access
+fund_agent_if_low(agent2.wallet.address())
```
jrriehl, line 16:
```suggestion
```

python/src/uagents/experimental/dialogues/__init__.py
```diff
@@ -0,0 +1,524 @@
...
+                        "model": edge.model.__name__ if edge.model else None,
+                    }
+                    for edge in self._edges
+                ],
+            }
+            if self._nodes
+            else {}
+        )
+
+    def _build_graph(self) -> dict[str, List[str]]:
```
jrriehl, line 239:
```suggestion
    def _build_graph(self) -> Dict[str, List[str]]:
```

python/src/uagents/agent.py
```diff
@@ -961,7 +963,52 @@ async def _process_message_queue(self):
...
```
jrriehl, line 1009: Can we add this logic to the `Dialogue` class instead of putting it here in `Agent`?

python/src/uagents/experimental/dialogues/__init__.py
```diff
@@ -0,0 +1,524 @@
...
+        """
+        graph = {}
+        for node in self._nodes:
+            graph.setdefault(node.name, [])
+        for edge in self._edges:
+            if edge.parent:
+                graph.setdefault(edge.parent.name, []).append(edge.child.name)
+        return graph
+
+    def _build_rules(self) -> dict[str, List[str]]:
```
jrriehl, line 254:
```suggestion
    def _build_rules(self) -> Dict[str, List[str]]:
```

jrriehl (commented):
It's been a while since I looked at this, so I mostly wanted to refresh my memory. Sorry if I've made some of these comments previously and they were already resolved.

python/src/uagents/agent.py
jrriehl: I think this is one of the comments I made once before. It seems I was unconvinced. :smile:

Dacksus: This logic here and in context that is actually only relevant to dialogues was the main reason to consider a greater refactor of the `Protocol` <> `Agent` <> `Context` interaction / dependencies. Of course we could reduce the footprint here by moving some code to the `Dialogue` class, but originally we left this here to actually have this discussion

python/src/uagents/context.py
```diff
@@ -429,6 +444,50 @@ async def send_raw(
...
+                    session_id=current_session,
+                    message_type=message_name,
+                    sender=self.address,
+                    receiver=destination,
+                    content=json_message,
+                )
+                current_protocol.update_state(schema_digest, current_session)
+                self.logger.debug(
+                    f"update state to: {message_name} for session {current_session}"
+                )
```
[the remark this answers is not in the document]
Dacksus: see above comment

python/examples/17-stateful-communication/main.py
```diff
@@ -0,0 +1,215 @@
...
+
+@chitchat_dialogue1.on_continue_dialogue(ChitChatDialogueMessage)
+async def continue_chitchat(
+    ctx: Context,
+    sender: str,
+    msg: ChitChatDialogueMessage,
+):
+    # do something when the dialogue continues
+    ctx.logger.info(f"Received message: {msg}")
+    counter = agent1.storage.get(str(ctx.session)).get("message_counter")
```
zmezei, line 95: Instead of agent1 can we just simply use ctx?

python/examples/17-stateful-communication/main.py
```diff
@@ -0,0 +1,215 @@
...
+    if counter < 3:
+        await ctx.send(
+            sender,
+            ChitChatDialogueMessage(text=f"Hello again #{counter}!"),
+        )
+        counter += 1
+        agent1.storage.set(str(ctx.session), {"message_counter": counter})
```
zmezei, line 102: The same as above, instead of `agent1` can we just simply use `ctx`?

Archento: fixed, the file has been removed.

python/examples/17-stateful-communication/main.py
Archento: fixed, the file has been removed.

python/src/uagents/agent.py
Archento: Exactly this has to do with the bigger refactor. Ideally we would try to put this code snippet in the `Dialogue` class but this is one of two instances where we need to directly tap into the uagents core mechanics so I don't see a way around that at the moment.

jrriehl (dismissed):
Looks good to me!

jrriehl (approved):

PR #189, opened 2023-10-19
fix: mailbox config

Alejandro-Morales:


python/src/uagents/config.py
```diff
@@ -92,7 +92,7 @@ def parse_agentverse_config(
     if "://" in base_url:
         protocol, base_url = base_url.split("://")
     protocol = protocol_override or protocol or "https"
-    http_prefix = "https" if protocol in {"wss", "https"} else "http"
+    http_prefix = "https" if protocol == "https" else "http"
```
jrriehl, line 95: I can we can leave this the way it was to make it easier to reactivate if we want.

python/src/uagents/mailbox.py
```diff
@@ -49,7 +49,7 @@ def protocol(self):
         """
         Property to access the protocol of the mailbox server.
 
-        Returns: The protocol of the mailbox server {ws, wss, http, https}.
+        Returns: The protocol of the mailbox server {http, https}.
```
jrriehl, line 52: Suggest reverting this change too.

python/tests/test_config.py
```diff
@@ -72,7 +72,7 @@
         "api_key": None,
         "base_url": "some_url",
         "protocol": "wss",
-        "http_prefix": "https",
+        "http_prefix": "http",
```
jrriehl, line 75: I think this is wrong, should be `https`.

python/tests/test_config.py
```diff
@@ -86,7 +86,7 @@
         "api_key": "api_key",
         "base_url": "agentverse.ai",
         "protocol": "wss",
-        "http_prefix": "https",
+        "http_prefix": "http",
```
jrriehl, line 89: This should also be `https`.

jrriehl (approved):
LGTM

PR #197, opened 2023-11-02
feat: automate faucet claims for test agents

jrriehl:
Check for funds before registration and do faucet claim if agent is on testnet.


python/src/uagents/agent.py
```diff
@@ -512,10 +515,18 @@ async def register(self):
...
                 )
-                return
+                if self._test:
+                    add_testnet_funds(str(self.wallet.address()))
+                    self._logger.info(
+                        f"Adding testnet funds to {self.wallet.address()}"
+                    )
+                else:
+                    self._logger.info(
+                        f"Fund using wallet address: {self.wallet.address()}"
```
NindoK, line 527: I think "Fund using wallet address:" is confusing as phrase to suggest to fund that wallet. "Add funds by sending FET to current wallet address:"?

NindoK (approved):

NindoK (approved):

PR #199, opened 2023-11-07
refactor: to make exchange adapter

ejfitzgerald:
## Summary

Some code refactoring to make it easier to build standalone scripts to send exchange messages


python/src/uagents/context.py
```diff
@@ -349,7 +349,8 @@ async def experimental_broadcast(
         agents = self.get_agents_by_protocol(destination_protocol, limit=limit)
         if not agents:
             self.logger.error(f"No active agents found for: {destination_protocol}")
-            return
+            return []
```
ejfitzgerald, line 352: My pycharm type checker was not a fan of this @Archento please check that I am not breaking something

python/src/uagents/context.py
```diff
@@ -364,7 +365,7 @@ async def experimental_broadcast(
             ]
         )
         self.logger.debug(f"Sent {len(futures)} messages")
-        return futures
+        return list(futures)
```
ejfitzgerald, line 368: Likewise here @Archento I think that `asyncio.gather` returns a `tuple()` - can you and @jrriehl give this a double check

python/src/uagents/agent.py
```diff
@@ -182,7 +183,11 @@ def __init__(
             if resolve is not None
             else GlobalResolver(max_endpoints=max_resolver_endpoints)
         )
-        self._loop = asyncio.get_event_loop_policy().get_event_loop()
+
+        if loop is not None:
+            self._loop = loop
```
qati, line 188: Nice!

python/src/uagents/context.py
```diff
@@ -457,12 +458,38 @@ async def send_raw(
...
+            schema_digest,
+            self.get_message_protocol(schema_digest),
+            json_message,
+            logger=self._logger,
+            timeout=timeout,
+            session_id=self._session,
+        )
+
+    @staticmethod
+    async def send_raw_exchange_envelope(
```
qati, line 474: Nice! If this could be released as 0.8.1 would be nice

jrriehl (approved):
Yeah, makes sense to me. I partly wonder if this should have been the original `send_raw` function, but I guess it's better to leave that in as well so as not to break anything.

python/src/uagents/context.py
Archento: I just checked the example with 0..2 responses and all were lists.
According to the [docs](https://docs.python.org/3/library/asyncio-task.html#running-tasks-concurrently) all successful awaitables will be aggregated to a list and as we always return with a `MsgStatus` this should therefore also always be a list meaning that we shouldn't need the explicit casting although it doesn't hurt either.

Archento (dismissed):

python/src/uagents/context.py
jrriehl: Reverting this change due to @Archento's findings.

jrriehl (approved):

PR #203, opened 2023-11-16
update api_key to mailbox_key

Alejandro-Morales:


python/src/uagents/config.py
```diff
@@ -75,31 +75,31 @@ def parse_agentverse_config(
     Returns:
         Dict[str, str]: The parsed agentverse configuration.
     """
-    api_key = None
+    mailbox_key = None
```
dominic22, line 78: please use the name `agent_mailbox_key`

jrriehl:
Moved to here: https://github.com/fetchai/uAgents/pull/204.

PR #215, opened 2024-01-23
feat: updates to match aname contract-V2

Alejandro-Morales:
This PR performs the needed updates to match `ANAME v2.0` contract with uAgents.
Now funds are required to register a domain, by default we set the registration time= 24 hours.
There are 2 main scenarios when attempting to register a domain:

- If the domain is available it will be registered and update the record with the agent address
- If the domain is not available (already registered and active) but the current wallet is the owner of the domain it will only update the record with the agent address
- If the domain is not available and the current wallet is not the owner of the domain, no registration will be attempted since another wallet owns the domain

For the name registration example (examples/13-agent-name-service) domain `example.agent` has been registered for around 3 months for "registration test wallet". So subdomains like `bob.example.agent` can be registered.

### DIFF
--- a/python/src/uagents/config.py
+++ b/python/src/uagents/config.py
@@ -25,7 +25,7 @@ MAINNET_CONTRACT_NAME_SERVICE = (
     "fetch1479lwv5vy8skute5cycuz727e55spkhxut0valrcm38x9caa2x8q99ef0q"
 )
 TESTNET_CONTRACT_NAME_SERVICE = (
-    "fetch1mxz8kn3l5ksaftx8a9pj9a6prpzk2uhxnqdkwuqvuh37tw80xu6qges77l"
+    "fetch1kewgfwxwtuxcnppr547wj6sd0e5fkckyp48dazsh89hll59epgpspmh0tn"
 )
 REGISTRATION_FEE = 500000000000000000
 REGISTRATION_DENOM = "atestfet"
--- a/python/src/uagents/config.py
+++ b/python/src/uagents/config.py
@@ -25,7 +25,7 @@ MAINNET_CONTRACT_NAME_SERVICE = (
     "fetch1479lwv5vy8skute5cycuz727e55spkhxut0valrcm38x9caa2x8q99ef0q"
 )
 TESTNET_CONTRACT_NAME_SERVICE = (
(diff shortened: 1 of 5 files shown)


python/src/uagents/network.py
```diff
@@ -409,28 +467,45 @@ async def register(
...
+        records = parse_record_config(agent_records)
+        agent_addresses = [val.get("address") for val in records]
+
+        for agent_address in agent_addresses:
+            if not get_almanac_contract(chain_id == "dorado-1").is_registered(
+                agent_address
+            ):
+                logger.warning(
+                    f"Address {agent_address} needs to be registered in almanac contract "
+                    f"to be registered in a domain"
```
jrriehl, line 486: Isn't a plus `+` needed here?
```suggestion
                    + "to be registered in a domain"
```

python/src/uagents/config.py
```diff
@@ -25,7 +25,7 @@
     "fetch1479lwv5vy8skute5cycuz727e55spkhxut0valrcm38x9caa2x8q99ef0q"
 )
 TESTNET_CONTRACT_NAME_SERVICE = (
-    "fetch1mxz8kn3l5ksaftx8a9pj9a6prpzk2uhxnqdkwuqvuh37tw80xu6qges77l"
+    "fetch1kewgfwxwtuxcnppr547wj6sd0e5fkckyp48dazsh89hll59epgpspmh0tn"
```
jrriehl, line 28: Let's revert this back to the main contract in preparation for release.

jrriehl (dismissed):
Nice one! Once we migrate the v2 contract to the main address, we can rerun the tests again and they should hopefully pass.

python/src/uagents/network.py
```diff
@@ -482,8 +482,9 @@ async def register(
                 agent_address
             ):
                 logger.warning(
-                    f"Address {agent_address} needs to be registered in almanac contract "
-                    + "to be registered in a domain"
+                    "Address %s needs to be registered in almanac contract "
```
jrriehl, line 485: OK, haven't used this formatting before, LGTM.

jrriehl (approved):
LGTM

PR #228, opened 2024-02-16
Feat: Live scores integration and Gemini AI integration

dpdrabla:
Adding live sports score integration


integrations/gemini_AI_integration/src/agents/user.py
```diff
@@ -0,0 +1,39 @@
+# Importing necessary libraries from uagents package
+from uagents import Agent, Context
+from uagents.setup import fund_agent_if_low
+from uagents import Model
+
+# Defining a model for messages
+class Message(Model):
+    message: str
+
+# Specifying the address of the restaurant agent
```
devjsc, line 10: comment seems misplaced

devjsc (changes requested):

integrations/gemini_AI_integration/src/agents/user.py
dpdrabla: comments corrected.

PR #232, opened 2024-02-20
feat: allow setting session_id in ctx.send

jrriehl:
Allow users to overwrite the session id by setting it in `ctx.send()`. This means that an agent can send messages for a previous session and not necessarily the one given by the incoming envelope.

This can be useful for example when information needs to be collected from other agents before responding to a request.


jrriehl:
Agreed, closing this.

PR #235, opened 2024-02-22
Feat/add mobility agents on mobility integrations

pratrivedi:


devjsc (changes requested):
Each agent should be it's own integration: 

- car service 
- car wash 
- geocode 

etc

PR #250, opened 2024-03-05
chore: rename broadcast method

Archento:
experimental -> official

### DIFF
--- a/python/examples/14-broadcast/main.py
+++ b/python/examples/14-broadcast/main.py
@@ -35,7 +35,7 @@ bob.include(proto)
 # let charles send the message to all agents supporting the protocol
 @charles.on_interval(period=5)
 async def say_hello(ctx: Context):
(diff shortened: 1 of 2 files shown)


jrriehl (commented):
Should we maintain `experimental_broadcast` as a simple wrapper for `broadcast` so as not to break code that is currently using this? Maybe with a deprecation warning?

Archento:
Although I agree in general with that approach, I would argue that we don't have that much of a user base currently that actively uses broadcast.
If we were to add a stub function acting as a wrapper we need to revisit this topic some time later and remove that function manually. I'm fine either way but the wrapper introduces more work so I'd prefer leaving it out

Archento:
Screenshot 2024-03-06 at 10 46 48

works like a charm

jrriehl (approved):
LGTM

PR #252, opened 2024-03-05
chore: restructure linting & formatting

Archento:
- changed `black` and `pylint` to `ruff` in poetry environment
- changed linting & formatting CI workflow to use ruff
- introduced `pre-commit` to local dev workflow for automatic linting&formatting before a commit
- applied isort (which apparently hasn't happened so far)
- applied a couple of small manual changes that were raised by ruff

I've run all tests and triple-checked the entire repo but please check again if possible for any braking changes.


Archento:
The `python/tests/test_config.py` test crashes in Python3.9 due to the `zip()` method not supporting the `strict` keyword argument. We need to be careful how we set the tests up, since it was not the unit test that failed but the test code being invoked by all Python distributions in our matrix.

Archento:
Based on the current runs this has been a 1/3 reduction in time needed for the linting&formatting workflow (~55s down to ~35s) + elimination of every workflow trigger based on a separate "I forgot linting & formatting" commit. 🎉

PR #259, opened 2024-03-13
feat: stateful agent

Archento:
Based on feedback from the community and the opened discussion #249.

Adds stateful behaviour to agents by providing the following capabilities without breaking backwards compatibility:
- add agent interface to get/set the current agent state as `str`
- add desired state as `str` to `.on_interval()` handler to only trigger that function based on the current agent state
- add desired state as `str` to `.on_message()` handler to only react to messages based on the current agent state
- decorators without the explicit mention of a state will act as usual

If desired we can add an example of state management


python/src/uagents/agent.py
```diff
@@ -1020,6 +1059,14 @@ async def _process_message_queue(self):
                     )
                     continue
 
+            if state and state != self.state:
+                await _send_error_message(
+                    context,
+                    sender,
+                    ErrorMessage(error="Agent not in correct state to handle message"),
```
Dacksus, line 1066: Should we support custom error messages the developer can provide with the decorator?

jrriehl: Currently if an agent has no handler for a message, it is simply ignored, so I kind of think we should treat being in wrong state to handle a message the same as not having a handler for that message. So for consistency with the current setup, we would ignore that case. However, I can see the case for sending error messages back in both cases, because at the moment I don't think the sending agent gets any feedback when sending an invalid or incorrect message to an agent.

jrriehl (commented):
Overall I like the idea and the implementation looks good and backwards compatible from the user's side. But given that this is a fundamental change to the framework, let's leave this open for a while longer for others to weigh in on it.

Archento:
Closing this PR as requirements have changed and the current framework is better kept lean.
If necessary the feature will be picked up at a later date.

PR #275, opened 2024-03-16
Pangea Pioneers Spoonacular API Agent

Ease-at-Bhumesh:


devjsc:
Project did not use uAgents in a way we'd expect for integrations. Great idea, just needed to use some agent communication

Ease-at-Bhumesh:
> Project did not use uAgents in a way we'd expect for integrations. Great idea, just needed to use some agent communication
Could you please clarify whether you're referring to communication among multiple agents or aspects related to user interface design? Additionally, could you let us know if it's possible for us to submit a pull request with some improvements, as we're quite interested in doing so?

PR #290, opened 2024-03-20
ci(core): add PR title linting to GH ci

Archento:


.github/workflows/pr-title-linting.yml
```diff
@@ -0,0 +1,33 @@
...
+            chore
+            docs
+            feat
+            fix
+            refactor
+            style
+            perf
+            ci
+            test
+            revert
```
jrriehl, line 27: Just a couple more suggestions:
```suggestion
            revert
            integration
            misc
```

jrriehl: Oh I see you made it a scope.

Archento: yes, I think that makes sense in the current structure :)

jrriehl (approved):
LGTM

PR #297, opened 2024-03-24
feat(integration): AI-Engine: Add dialogue support

qati:


NindoK (approved):

NindoK (approved):

integrations/fetch-ai-engine/src/ai_engine/dialogue.py
```diff
@@ -0,0 +1,97 @@
...
+
+    # is the node a terminal node
+    terminal: bool = False
+
+
+class EdgeMetadata(BaseModel):
+    """Metadata for the edges"""
+
+    # what is the target of this transition, user means direct message to the user, ai, system involves AI Engine processing
+    target: Literal["user", "ai", "system"]
```
Dacksus, line 25: what's the difference between "ai" and "system"?

jrriehl (approved):
Looks good to me. I guess that some of this (like what's in `dialogue.py`) could be directly integrated into `uAgents.experimental.Dialogue`, but I'd say it's okay to get it in here first.

PR #304, opened 2024-04-02
feat(core): add default state support to dialogues

Archento:


python/examples/17-stateful-communication/Readme.md
```diff
@@ -1,14 +1,24 @@
...
 #### Define a pattern (generic dialogue without models)
+
 1. Define the intended behaviour / communication flow as a directed graph.
 2. Create a new class inheriting from `Dialogue`.
 3. Specify nodes and edges according to the graph.
 4. Specify a concrete transition decorator for each transition (=edge).
 
 See `chitchat.py` for an example.
 
+**Note:** You can design your graph to have a default state which represents the state that the agent is in when a dialogue with another agent hasn't been started yet. This state is not shared but exists for each potential dialogue instance with other agents. Since agents' interactions are message centric, this first / default state has no impact on the whole dialogue and is therefore optional.
```
jrriehl, line 12: I'm not really convinced that the default state should be optional. I would be inclined to omit this last sentence and no longer support edges without entry. 

Just a suggestion:
```suggestion
**Note:** The default state represents the state that the agent is in when a dialogue with another agent hasn't been started yet. This state is not shared but exists for each potential dialogue instance with other agents. The default state is included for completeness of the Dialogue graph and has no functional role .
```

Dacksus (dismissed):

Dacksus (approved):

PR #306, opened 2024-04-03
feat(core): add capability to define default behaviour to dialogue edges

Archento:
Adding the capability to define default behaviour to dialogue edges.

The example is modified so that when including the ChitChatDialogue all attempts of initiating a dialogue will be rejected by default and will only be allowed if you specifically write a handler that does so.

> This is just a minor example to demonstrate how we can effectively build complex dialogues that automate certain steps without the user having to modify/copy every individual step (which is part of the main benefits of dialogues from their inception)

### DIFF
--- a/python/src/uagents/agent.py
+++ b/python/src/uagents/agent.py
@@ -761,10 +761,11 @@ class Agent(Sink):
         self._interval_messages.update(protocol.interval_messages)
 
         for schema_digest in protocol.models:
-            if schema_digest in self._models:
-                raise RuntimeError("Unable to register duplicate model")
-            if schema_digest in self._signed_message_handlers:
-                raise RuntimeError("Unable to register duplicate message handler")
+            if not hasattr(protocol, "rules"):
+                if schema_digest in self._models:
+                    raise RuntimeError("Unable to register duplicate model")
+                if schema_digest in self._signed_message_handlers:
+                    raise RuntimeError("Unable to register duplicate message handler")
             if schema_digest in protocol.signed_message_handlers:
                 self._signed_message_handlers[schema_digest] = (
                     protocol.signed_message_handlers[schema_digest]
--- a/python/src/uagents/experimental/dialogues/__init__.py
+++ b/python/src/uagents/experimental/dialogues/__init__.py
@@ -40,15 +40,13 @@ class Edge:
         description: str,
         parent: Optional[Node],  # tail
         child: Node,  # head
-        model: Optional[Type[Model]] = None,
-        func: MessageCallback = lambda *args, **kwargs: None,
     ) -> None:
         self.name = name
         self.description = description
         self.parent = parent
         self.child = child
-        self._model = model
-        self._func = func
+        self._model = None
+        self._func = None
 
     @property
     def model(self) -> Optional[Type[Model]]:
@@ -57,6 +55,7 @@ class Edge:
 
     @model.setter
     def model(self, model: Type[Model]) -> None:
+        """Set the message model type for the edge."""
         self._model = model
 
     @property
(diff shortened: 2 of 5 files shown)


Archento:
While we are able to provide default behaviour to some or all of the graphs edges, this feature also allows a graph designer to only expose handlers for some of the interactions effectively 'hiding' some steps from the user of the dialogue.
One of the use-cases could be a re envisioned token payment dialogue in which two agents interact and agree on a token price for a specific service but the actual exchange of the wallet address and token payment can be 'automated' within the dialogue.

python/src/uagents/agent.py
```diff
@@ -761,10 +761,11 @@ def include(self, protocol: Protocol, publish_manifest: Optional[bool] = False):
         self._interval_messages.update(protocol.interval_messages)
 
         for schema_digest in protocol.models:
-            if schema_digest in self._models:
-                raise RuntimeError("Unable to register duplicate model")
-            if schema_digest in self._signed_message_handlers:
-                raise RuntimeError("Unable to register duplicate message handler")
+            if not hasattr(protocol, "rules"):
```
Dacksus, line 764: I still don't see the need for this. The overwriting of the default behavior by the decorator happens before `include` is called, so this is not triggered. Once we are here, there will be only one model entry with the function registered by the latest decorator (or latest call of `set_default_behaviour`). It runs perfectly for me without this line.

There is, however, one issue that is also not caught by this: If we overwrite the transition with a decorator specifying another `Model` than what was used in `set_default_behaviour`, (e.g., `@chitchat_dialogue.on_init_session(RejectChitChatDialogue)` the agent will crash, as this is a yet undefined state. We would either need to enforce that if a default handler has been set, overwriting is only possible using the same `Model` or make sure to completely remove the earlier  registration.

python/src/uagents/experimental/dialogues/__init__.py
```diff
@@ -300,7 +300,7 @@ def get_current_state(self, session_id: UUID) -> str:
     def _auto_add_message_handler(self) -> None:
         """Automatically add message handlers for edges with models."""
         for edge in self._edges:
-            if edge.model:
+            if edge.model and edge.func:
                 model_digest = Model.build_schema_digest(edge.model)
                 self._models[model_digest] = edge.model
                 self._signed_message_handlers[model_digest] = edge.func
```
Dacksus, line 306:
```suggestion
                self._add_message_handler(edge.model, edge.func, None, False)
```

Dacksus (changes requested):

python/src/uagents/agent.py
Archento: You are right, this must have been an issue that was already handled / fixed by the other commits. I definitely remember having to add this line to make it work but it doesn't seem to be necessary any more.

As for the other point, it makes sense to restrict overwrites of edges' handlers to the same message Models as both participants need to support the same models (due to the digests in the manifest) to properly communicate. Ultimately this comes down to how and what we actually put into the manifest going forward.

Dacksus: With the current implementation of manifests this would eliminate the possibility of defining a general pattern that can be instantiated with different models, as soon as a default behavior is defined. Conceptually, it would be nicer to still keep patterns general and provide default behavior with concrete dialogues. However, I'm not sure if the current architecture really supports this, since the distinction between the logical `Pattern` and `Dialogue` is only based on whether a `Dialogue` is instantiated or not...

I guess this the long way of saying, I agree it should just be restricted in this version and needs to be reevaluated in the bigger picture.

Archento: If the models are used as a general purpose vehicle this can still work. I'm thinking about a similar approach to what Attila has done with the Metadata where the contents of the message can still be overloaded to contain custom information while retaining the connection of message<->interaction

jrriehl (approved):
Makes sense to me.

python/src/uagents/agent.py
Archento: Going to close this conversation since we take it to a separate meeting.

jrriehl (approved):

Dacksus (approved):

PR #308, opened 2024-04-05
ci(core): upgrade CI actions versions

Archento:
This PR aims to fix:
- deprecation warnings regarding actions that use NodeJS 16
- use latest version of Ruff for python linting and formatting
- remove Python3.8 from the readme and CI as it is not part of the pyproject.toml for a while now so we didn't support it explicitly

Annotations from the GH actions:
Screenshot 2024-04-05 at 16 26 39


jrriehl (approved):
LGTM

PR #309, opened 2024-04-05
refactor(example): examples overhaul

Alejandro-Morales:


jrriehl (commented):
Looking good! Some suggestions:
- It's prettly light on core examples so far. I'd suggest adding at least a basic protocol example here to make sure that is included in the core
- We discussed adding README.md files to the examples to briefly describe the purpose of each one

python/examples/network/01-bureau-interaction/wallet-messaging.py
jrriehl, line 1: Can we move the wallet messaging example (without bureau) into core? This is also supported in Agentverse.

python/examples/network/01-bureau-interaction/README.md
jrriehl, line 1: @archento had an idea for an example that highlighted some useful usage of the `Bureau` as a local agent scaling / load-balancing solution. It might be nice to include that here instead of another alice and bob example.

Dacksus:
examples overhaul

- [ ] We should also add a readme to the tortoise integration
- [ ] We should add a brief top-level readme to the examples directory, explaining the separation between agent-core and network, how to navigate/use it

To improve maintainability (no redundancy, single source for documentation/definitions) I would suggest the following:
- Only include example logic specific descriptions and explanations in the example's readme
- put all "general" documentation into (a) separate file(s)
- Link to any "general" description/documentation that is needed for an example (e.g., address, seed, bureau etc. basically everything that also should have a proper docstring)
- If we switch from markdown to asciidocs we can even do includes to compile standalone files without linking and without copy-and-paste redundancies
- eventually we should maintain valid, useful and up-to-date docstrings, which should be used to populate any base documentation (e.g., in example readmes or for general api documentation) and additional documentation would only be needed for concepts, architecture descriptions, examples, etc.

jrriehl:
Good points @Dacksus. Definitely agree with both of the first tasks, and for these:

> - Only include example logic specific descriptions and explanations in the example's readme
> - put all "general" documentation into (a) separate file(s)
> - Link to any "general" description/documentation that is needed for an example (e.g., address, seed, bureau etc. basically everything that also should have a proper docstring)

Are you suggesting to provide links like this for all the relevant objects / concepts: https://fetch.ai/docs/references/uagents/uagents-api/context?

PR #312, opened 2024-04-12
feat(core,example): better support for sending synchronous messages

jrriehl:
Currently it is not easy to send a signed synchronous message. Unsigned sync messages can be sent via the `query` function in `query.py`, but signed sync messages require a lot of extra work.

This PR allows sending messages synchronously with `Context.send(..., sync=True)` and a standalone function `send_sync_message` with a demonstration in the examples.


ejfitzgerald (approved):

python/src/uagents/context.py
```diff
@@ -634,6 +686,22 @@ async def send_raw_exchange_envelope(
...
+    @staticmethod
+    async def dispatch_sync_response_envelope(env: Envelope) -> MsgStatus:
+        await dispatcher.dispatch(
+            env.sender,
+            env.target,
+            env.schema_digest,
+            env.decode_payload(),
+            env.session,
+        )
+        return MsgStatus(
```
Dacksus, line 698: This means that the `MsgStatus` for the sent message is delivered/received after the actual response has already been handled?

I guess it's not of any relevance, just trying to follow this sync async flow.

jrriehl: Right, so the message is going out to the destination as a sync response and the status is going back to the sender as a successful sync delivery.

Archento (approved):
All good from my side. Nice work!

PR #316, opened 2024-04-13
Movie Database Integration

darthtakt:
Project Overview :
This project takes the title of the movie from the client as it input and provides information : movie rating & no. of reviews from the designated platform.

Functionalities :
Explained below are the functions of different set of codes involved in the project : 
1. User : Used to receive the title from the client & convert the title into movie-id, which serves as the input for the agents. 
2. Agent : Used  to provide the movie rating &  no. of votes extracted using the movie-id.
3. Model : Provides Message & Response function for the agents.
4. Main : Adds all agents to a common bureau.


devjsc:
Main import agent1, and agent2 but these files do not exist in the project. PR also deletes another integration. Please fix these issues and re-open PR when you're ready and we'll review again.

PR #321, opened 2024-04-13
Jsearch integration with uagents

Shruti27272727:


gautamgambhir97:
we already have this integration:
https://github.com/fetchai/uAgents/tree/main/integrations/job-finder

PR #345, opened 2024-04-16
refactor(core): dialogue logic restructure

Archento:
- add dialogue starter method
- add conversation history interface
- completely remove dialogue logic from `agent.py` and `context.py`
- remove custom session methods


jrriehl (approved):
Love it! :star: :star: :star: :star: :star:

PR #349, opened 2024-04-19
refactor(core): context stage 1

Archento:
This is the first iteration of `Context` refactoring where all agent related properties within the context are clustered to one `ctx.agent` property. This is possible with the new `AgentRepresentation` class and enables lean passthrough of necessary properties without exposing unnecessary interfaces.

context_refactor-Page-3


jrriehl (commented):
Looks good to me. Since there are a few breaking changes here, I think we just need to compile a list of them somewhere that we can include in release notes, docs, etc.

Archento:
PR will be closed in favour of #363 as all changes from this PR are included there as well.

PR #357, opened 2024-04-21
feat(core): native rest endpoints for agents

ejfitzgerald:
## Summary

Initial concept of creating native rest endpoint support for agents.

### Missing features

- [x] Add support for Bureau for rest endpoints
- [ ] ~~Agentverse hosting updates for rest endpoints~~
- [x] Adding support for agents who are using the Mailbox
- [x] Probably a whole load of additional things too 😁 
- [x] Add tests

**@Archento Note**:  We'll skip the Agentverse dependency for this PR to get this feature finished from the uagents perspective and will most likely follow that up later.


devjsc:
🎉

python/src/uagents/agent.py
```diff
@@ -996,11 +1061,13 @@ async def run_async(self):
 
         """
         await self.setup()
+
+        tasks = [self._server.serve()]
```
jrriehl, line 1065: Should we check for any registered REST handlers before running the server? Seems like this could save resources for a mailbox agent that has opted out of handling any rest endpoints, but not sure how significant.

python/src/uagents/agent.py
```diff
@@ -996,11 +1061,13 @@ async def run_async(self):
...
+        if self._use_mailbox and self._mailbox_client is not None:
```
jrriehl, line 1066: Should we check for any registered REST handlers before running the server? Seems like this could save resources for a mailbox agent that has opted out of handling any rest endpoints, but not sure how significant.

python/tests/test_rest.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+class Response(Model):
+    text: str
+
+
+agent = Agent(name="alice")
+
+
+@pytest.mark.order(1)
```
jrriehl, line 22: :+1:

python/src/uagents/agent.py
Archento: Good point. Two things come to mind though:
- we have to consider if the resource benefit is worth shutting down the server for an agent that only communicates via mailbox
- if the agent uses the inspector it will need a server anyway to provide the respective interfaces and currently we don't have an opt-out mechanism in place

Archento: I refactored the code to:
- Allow the bureau to have agents with the same rest endpoints (if multiple endpoints are found it needs the "x-uagents-address" header to distinguish where to route to)
- Deactivate the server if mailbox is used and no rest endpoints are registered

jrriehl (approved):

PR #359, opened 2024-04-22
refactor(integration): add fetch and community integration categories

jrriehl:
Categorize all integrations as `fetch` official or `community` contributions.


Archento:
Is this still needed?

PR #363, opened 2024-05-07
refactor(core): context stage 2

Archento:
List of changes:
- includes (and depends on) changes from #349
- add communication module for handling sending of messages
- add `Dispenser` class in communication module
- add new background task in `agent.py` for message dispensing
- add new `Context` interface / Abstract Base Class to be used by the handlers
- split existing Context into two types:
  - `InternalContext`: for supporting pro-active behaviour (on_event, on_interval)
  - `ExternalContext`: for supporting re-active behaviour (on_message)
- drop protocol discovery feature
- remove duplicate code
- remove background_tasks set


python/src/uagents/agent.py
```diff
@@ -241,12 +337,18 @@ def __init__(
...
         self._message_queue = asyncio.Queue()
         self._on_startup = []
         self._on_shutdown = []
         self._test = test
         self._version = version or "0.1.0"
 
+        # configure the dispenser
+        self._dispenser.configure(
+            resolver=self._resolver,
+        )
```
jrriehl, line 350: Since the dispenser depends on a resolver, unlike the dispatcher, should we simply create it here instead of modifying this global which would affect all other usage of it?

python/src/uagents/config.py
```diff
@@ -115,3 +117,24 @@ def get_logger(logger_name, level=logging.INFO):
...
+    Create a task and add it to the agents event loop and task list.
+
+    Args:
+        loop (asyncio.AbstractEventLoop): Agent event loop.
+        task_list (Set[asyncio.Task]): Agent internal task list.
+        func (Callable): Function to be executed as a task.
+    """
+    task = loop.create_task(func)
+    task_list.add(task)
+    task.add_done_callback(task_list.discard)
```
jrriehl, line 140: These would seem to be a better fit for a new `utils.py` or `helpers.py` module.

python/src/uagents/dispatch.py
```diff
@@ -19,7 +19,7 @@ async def handle_message(
 
 class Dispatcher:
     """
-    Dispatches messages to sinks.
+    Dispatches messages internally to sinks.
```
jrriehl, line 22:
```suggestion
    Dispatches incoming messages to internal sinks.
```

python/src/uagents/envelope.py
```diff
@@ -33,7 +33,7 @@ class Envelope(BaseModel):
     sender: str
     target: str
     session: UUID4
-    schema_digest: str = Field(alias="protocol")
+    schema_digest: str = Field(alias="protocol")  # ???
```
jrriehl, line 36: This field was initially called `protocol` unfortunately, and this alias was created to preserve backwards compatibility, but this needs to be revisited.

python/src/uagents/communication.py
```diff
@@ -0,0 +1,268 @@
...
+                    endpoint,
+                    headers=headers,
+                    data=envelope.json(),
+                ) as resp:
+                    success = resp.status == 200
+                    if success:
+                        if sync:
+                            return await dispatch_sync_response_envelope(
+                                Envelope.parse_obj(await resp.json())
+                            )
```
jrriehl, line 167: This should only dispatch the envelope if there is an agent to dispatch to, which is not always the case for this standalone function (see https://github.com/fetchai/uAgents/blob/main/python/examples/08-local-network-interaction/sync_sender.py). Otherwise it should return the envelope itself.

python/src/uagents/communication.py
```diff
@@ -0,0 +1,268 @@
...
+        destination=destination,
+        endpoint="",
+    )
+
+
+async def send_exchange_envelope(
+    envelope: Envelope,
+    resolver: Optional[Resolver] = None,
+    sync: bool = False,
+) -> Union[MsgStatus, Envelope]:
```
jrriehl, line 126: The destination of the envelope actually depends on the output of the resolver, so the envelope may need to be created inside this function.

python/src/uagents/communication.py
```diff
@@ -0,0 +1,268 @@
...
+    Returns:
+        Union[Model, JsonStr, MsgStatus]: On success, if the response type is provided, the response
+        message is returned with that type. Otherwise, the JSON message is returned. On failure, a
+        message status is returned.
+    """
+    if sender is None:
+        sender = Identity.generate()
+
+    # TODO: This is not quite right as this should really be resolved by the resolver
+    _, _, destination_address = parse_identifier(destination)
```
jrriehl, line 242: I only added this to make things work when the envelope is created before using the resolver.

python/src/uagents/context.py
```diff
@@ -478,105 +376,56 @@ async def send_raw(
...
+
         # Extract address from destination agent identifier if present
         _, _, destination_address = parse_identifier(destination)
 
         if destination_address:
             # Handle local dispatch of messages
             if dispatcher.contains(destination_address):
-                await dispatcher.dispatch(
-                    self.address,
+                return dispatch_local_message(
```
jrriehl, line 405:
```suggestion
                return await dispatch_local_message(
```

python/src/uagents/context.py
```diff
@@ -585,136 +434,165 @@ async def send_raw_exchange_envelope(
...
-            status=DeliveryStatus.FAILED,
-            detail="Message delivery failed",
+            status=DeliveryStatus.QUEUED,
+            detail="Envelope successfully queued",
             destination=destination,
             endpoint="",
         )
+        # At this point we need the result of the send_raw_exchange_envelope
+        # but the result will be async so we need another way to access the
+        # return value
```
jrriehl, line 458: Yes, indeed. How about adding a `Future` to self._outbound_messages` which the dispenser resolves to the corresponding `MsgStatus`, and then awaiting that here?

jrriehl (commented):
A few remaining subtleties to address, but overall this looks great.

python/src/uagents/config.py
Archento: Good idea, moved the two functions over to `utils.py`

python/src/uagents/envelope.py
Archento: Created issue #365

python/src/uagents/context.py
Archento: done

python/src/uagents/agent.py
Archento: I was thinking the exact same thing but somehow still decided to have it globally available. Creating it here would mean to pass it down to the Context and my initial goal was to keep the context as small as possible.

If there are no more big changes it would make sense to create the dispenser here (since it is intended as a 1 per one agent basis anyway)

python/src/uagents/communication.py
Archento: I wasn't aware of that and unfortunately the naming of `destination` and `destination_address` was a bit misleading. In all of my tests those where identical but I have to admit, I didn't try the ANS.
But this means that the resolver logic needs to be put back into the context (which I'm fine with) as the goal here should be to have the `context.send()` method create the `Envelope` and the actual sending happens elsewhere.

jrriehl: That works, but in that case, we might want to create a standalone wrapper for `send_exchange_envelope` that constructs the envelope similar to the way this function does: https://github.com/fetchai/uAgents/blob/main/python/src/uagents/context.py#L534-L548.

python/src/uagents/context.py
```diff
@@ -437,39 +459,39 @@ async def send_raw(
...
-            endpoint="",
-        )
-        # At this point we need the result of the send_raw_exchange_envelope
-        # but the result will be async so we need another way to access the
-        # return value
+        self._queue_envelope(env, endpoints, fut, sync)
 
-        # if isinstance(result, Envelope):
-        #     return await dispatch_sync_response_envelope(result)
+        result = await fut
```
jrriehl, line 474: Should we use `asyncio.wait_for(fut, timeout)` here?

Archento: Agree, I had it in earlier but reverted to the current state.
My reasoning was that the `send_exchange_envelope` method always returns a form of `MsgStatus` if something goes wrong but it wouldn't hurt to add a timeout fallback here.

python/src/uagents/agent.py
```diff
@@ -92,6 +106,87 @@ async def _send_error_message(ctx: Context, destination: str, msg: ErrorMessage)
...
+
+    @property
+    def identifier(self) -> str:
+        """
+        Get the address of the agent used for communication including the network prefix.
+
+        Returns:
+            str: The agent's address and network prefix.
+        """
+        return TESTNET_PREFIX + "://" + self._address
```
Dacksus, line 175: `prefix = TESTNET_PREFIX if self._test else MAINNET_PREFIX`
`return prefix + "://" + self._identity.address`
?

python/src/uagents/context.py
```diff
@@ -49,275 +54,277 @@
...
+
+    @property
+    def storage(self) -> KeyValueStore:
+        return self._storage
+
+    @property
+    def ledger(self) -> LedgerClient:
+        return self._ledger
 
-    def update_protocols(self, protocol: Protocol) -> None:
```
Dacksus, line 291: just to double check: Dialogues are still working with this removed?

python/src/uagents/context.py
```diff
@@ -329,249 +336,136 @@ def get_agents_by_protocol(
...
+        destination: str,
+        message: Model,
+        sync: Optional[bool] = False,
+        timeout: Optional[int] = DEFAULT_ENVELOPE_TIMEOUT_SECONDS,
+    ) -> MsgStatus:
+        """
+        This is the pro-active send method which is used in on_event and
+        on_interval methods. In these methods, interval messages are set but
+        we don't have access properties that are only necessary in re-active
+        contexts, like 'replies', 'message_received', or 'protocol'.
```
Dacksus, line 396: Mind to rephrase? I'm not really sure what this docstring is telling me and since `send` is a fairly popular method, documentation should be super clear and concise

python/src/uagents/context.py
```diff
@@ -329,249 +336,136 @@ def get_agents_by_protocol(
...
 
+        return await self.send_raw(
+            destination,
+            schema_digest,
+            message_body,
+            sync=sync,
+            timeout=timeout,
+        )
+
+    async def send_raw(
```
Dacksus, line 418: If we need to keep this method public, we should also add a docstring explaining when this should be used (or rather why the common developer should never need to use this)

python/src/uagents/context.py
```diff
@@ -585,136 +479,180 @@ async def send_raw_exchange_envelope(
...
 
-async def send_sync_message(
-    destination: str,
-    message: Model,
-    response_type: Type[Model] = None,
-    sender: Identity = None,
-    resolver: Resolver = None,
-    timeout: int = 30,
-) -> Union[Model, JsonStr, MsgStatus]:
+class ExternalContext(InternalContext):
```
Dacksus, line 544: I don't have a good suggestion yet, but I feel we could improve on the naming and description of these contexts. It should be clear what "Internal", "External", "reactive" and "proactive" etc means to both new developers as well as those that already worked with the previous solitary context

Dacksus (changes requested):

python/src/uagents/communication.py
```diff
@@ -0,0 +1,282 @@
...
+                        envelope=env,
+                        endpoints=endpoints,
+                        sync=sync,
+                    )
+                    response_future.set_result(result)
+                except Exception as err:
+                    LOGGER.error(f"Failed to send envelope: {err}")
+                finally:  # sending an envelope is only tried once
+                    self._envelopes.remove((env, endpoints, response_future, sync))
+            await asyncio.sleep(DEFAULT_DISPENSER_INTERVAL)
```
jrriehl, line 116: Not sure there is any need to sleep here. Why not just let this run continuously?

Archento: I started out with a timeout as I didn't want to unnecessarily take up resources. We can throw the whole part out as well as the config as that is currently not needed.

python/src/uagents/communication.py
```diff
@@ -0,0 +1,282 @@
...
+class Dispenser:
+    """
+    Dispenses messages externally.
+    """
+
+    def __init__(self):
+        self._envelopes: List[Tuple[Envelope, List[str], asyncio.Future, bool]] = []
+        self._timeout: float = DEFAULT_DISPENSER_INTERVAL
+
+    def configure(self, timeout: Optional[float] = None):
```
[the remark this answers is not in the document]
Archento: I'll throw out the config method as that is not needed once the timeout is gone.

python/src/uagents/communication.py
```diff
@@ -0,0 +1,282 @@
...
+    )
+    return MsgStatus(
+        status=DeliveryStatus.DELIVERED,
+        detail="Message dispatched locally",
+        destination=destination,
+        endpoint="",
+    )
+
+
+async def send_exchange_envelope(
```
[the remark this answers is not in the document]
Archento: The envelope can have async and sync responses, i.e. exchanges. At least I think that is the origin of the name.
We'll revisit that topic at a later stage, in addition to reevaluating the need for `send_raw`

python/src/uagents/communication.py
```diff
@@ -0,0 +1,282 @@
...
+            async with aiohttp.ClientSession() as session:
+                async with session.post(
+                    endpoint,
+                    headers=headers,
+                    data=envelope.json(),
+                ) as resp:
+                    success = resp.status == 200
+                    if success:
+                        if sync:
+                            return await dispatch_sync_response_envelope(
```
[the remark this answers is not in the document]
Archento: Good point, but that will be covered in a separate issue.

python/src/uagents/communication.py
```diff
@@ -0,0 +1,282 @@
...
+        sender=sender.address,
+        target=destination_address,
+        session=uuid.uuid4(),
+        schema_digest=Model.build_schema_digest(message),
+        expires=int(time()) + timeout,
+    )
+    env.encode_payload(message.json())
+    env.sign(sender.sign_digest)
+
+    response = await send_exchange_envelope(
```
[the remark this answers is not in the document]
Archento: The send_sync_message is a purposefully created separate method to simplify sending synchronous messages to agents. It is meant to somewhat replace on_query at some point but the whole topic  of method structuring needs a revisit at a later stage.

python/src/uagents/agent.py
Archento: Noted. We'll cover that when switching to main net as there are a couple more implications when putting that into the `AgentRepresentation`

python/src/uagents/context.py
Archento: Yes, the Dialogue examples have been thoroughly tested with the new context and this was only needed at the stage where Dialogues were more deeply embedded in the context itself.

python/src/uagents/context.py
Archento: This is an internal comment, the developer facing comments are located at the abstract base class.

python/src/uagents/context.py
Archento: You are absolutely right, the naming is essential but everything that is changed in this refactor is not visible to the developer. We purposely use the new Context Interface class to specifically design the usable aspects of the Context. Nothing that is changed in this PR should affect the user at all.

jrriehl (approved):
LGTM

PR #370, opened 2024-05-23
refactor(core): dialogue cleanup task handling

Archento:
provide an interface for the cleanup task interval (which acts as a multiplier for the timeout that is set per Dialogue instance), i.e. it affects how long the messages are seen as valid.

Setting it to 0 will disable the cleanup task altogether if you want to handle that separately.


jrriehl (approved):
LGTM

PR #371, opened 2024-05-24
feat(core): add persistent dialogue sessions

Archento:
The agent will be able to pick up existing sessions from storage that are still valid.
This enables you to restart an agent within an active dialogue and continue as before.


jrriehl (approved):
LGTM

PR #374, opened 2024-05-24
manifest refactor wip

Dacksus:
- adds separation of functional description and metadata of nodes and edges for dialogues
- adds separation of functional description and metadata of models for protocols
- cached protoco.digest to avoid unnecessary recomputation

manifest structure is robust (same manifest created for different orderings of the same protocol/dialogue), but digest computation is not robust yet as model digests are yet sensible to ordering, docstrings etc. -> Todo

### DIFF
--- a/python/src/uagents/experimental/dialogues/__init__.py
+++ b/python/src/uagents/experimental/dialogues/__init__.py
@@ -27,11 +27,13 @@ class Node:
         name: str,
         description: str,
         initial: bool = False,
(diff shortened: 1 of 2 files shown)


Dacksus:
updated model digestion (to use pydantic v2). -> Needs refactoring throughout the framework to update deprecated function calls (see tests)

beshr11 (approved):

PR #385, opened 2024-05-29
feat(core): add broadcast functionality to dialogues

Archento:
Also the following:
- add optional session id to MsgStatus
- fix broadcast method due to changed API
- refactor error messages in send_exchange_envelope method

It is now possible to use a protocol digest for the destination argument in `Dialogue.start_dialogue(...)` to start a dialogue with multiple agents at the same time.


Archento:
Do we need an example for this?

jrriehl (approved):
LGTM

jrriehl (approved):

PR #399, opened 2024-05-30
chore(core): fix typing issues

Archento:
## Proposed Changes

This PR addresses a lot of typing inconsistencies and aims to fix Pylance and Pyright warnings.
It also fixes fetching the correct protocol on creation of the ExternalContext within `_process_message_queue()` and the odd typo and missing doctoring.

## Types of changes
- [x] Documentation update.

## Checklist
  - [x] I have read the [CONTRIBUTING ↗️](/CONTRIBUTING.md) guide

### If applicable

  - [x] I have checked that code coverage does not decrease

### DIFF
--- a/python/src/uagents/context.py
+++ b/python/src/uagents/context.py
@@ -122,12 +122,23 @@ class Context(ABC):
         """
         pass
 
+    @property
+    @abstractmethod
+    def session(self) -> Union[uuid.UUID, None]:
+        """
+        Get the session UUID associated with the context.
+
+        Returns:
+            uuid.UUID: The session UUID.
+        """
+        pass
+
     @abstractmethod
     def get_agents_by_protocol(
         self,
         protocol_digest: str,
-        limit: Optional[int],
-        logger: Optional[logging.Logger],
+        limit: int = DEFAULT_SEARCH_LIMIT,
+        logger: Optional[logging.Logger] = None,
     ) -> List[str]:
         """Retrieve a list of agent addresses using a specific protocol digest.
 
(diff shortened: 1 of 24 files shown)


python/src/uagents/context.py
```diff
@@ -639,14 +655,12 @@ async def send(
...
-        if schema_digest != ERROR_MESSAGE_DIGEST and not self._is_valid_reply(
-            schema_digest
-        ):
+        if not self._is_valid_reply(schema_digest):
             log(
                 self.logger,
                 logging.ERROR,
-                f"Outgoing message {message_type} is not a valid reply"
-                f"to {self._message_received.message}",
+                f"Outgoing message '{message_type}' is not a valid reply"
```
jrriehl, line 662: Isn't it more informative to report the human readable message types here?

Archento: If you're referring to the `self._message_received` you are absolutely right. Unfortunately message_received only has the message body and the digest so we can't produce the type at this stage.
As for the `message_type` variable: I was trying to find a better solution than printing the object str representation but left it with `type()` because it was the easiest thing to do.

jrriehl: I see, so previously this was logging the whole message contents. It seems that we could easily have the `model_class`, but we just don't carry it through to here: https://github.com/fetchai/uAgents/blob/main/python/src/uagents/agent.py#L1021-L1041. I suppose we could add it as an optional field in the `MsgDigest` object, but not sure whether it's worth it just for logging this message.

jrriehl: We also don't validate the message contents against the model class before building the context.

Archento: Indeed, this is the only point at which the model would be used so I also don't think that it's worth putting the model in there. (In addition to the misleading class name)

Also, the validation of the message happens within the asgi module before being queued for pickup by the agent so I don't necessarily think that we need to validate again here.

jrriehl: I guess that the asgi module validates the Envelope instance while the agent validates the envelope's payload against the specific Model instance, but I agree that it's not worth a pre-validation of the message just for this error log.

jrriehl (approved):
LGTM

PR #435, opened 2024-07-02
chore(core): misc refactor for almanac api resolver

lrahmani:


python/src/uagents/resolver.py
```diff
@@ -293,9 +304,29 @@ async def resolve(self, destination: str) -> Tuple[Optional[str], List[str]]:
...
+            destination (str): The destination address to resolve.
+
+        Returns:
+            Tuple[Optional[str], List[str]]: The address and resolved endpoints.
+        """
+        address, endpoints = await self._api_resolve(destination)
+        return (
+            (address, endpoints)
+            if address is not None
+            else await self._almanac_contract_resolver.resolve(address)
```
Alejandro-Morales, line 328: Should be 
`else await self._almanac_contract_resolver.resolve(destination)`
Because address would be None

lrahmani: Good catch! That's a bug

lrahmani, line 328:
```suggestion
            else await self._almanac_contract_resolver.resolve(destination)
```

Alejandro-Morales (approved):

PR #440, opened 2024-07-03
fix(core): catch destination identifiers

Alejandro-Morales:
In this PR, we are updating the resolver to use destination instead of parsed_address. This change is necessary because destination contains all the information required by the resolver to determine whether to query testnet or mainnet, particularly when using AlmnacContractResolver and NameServiceResolver.


Archento (approved):
LGTM!

PR #451, opened 2024-07-09
fix(core): digest backwards compatibility

Archento:
Reverting back to the old way of schema generation to be compatible with current agents.

We will introduce a transitory period of supporting both schemes in the near future to provide a chance for users to properly migrate,


python/tests/test_model.py
```diff
@@ -25,6 +27,40 @@ class SuperImportantCheck(Model):
 
         self.assertEqual(result, target_digest, "Digest mismatch")
 
+    def test_calculate_nested_digest_backcompat(self):
```
qati, line 30: 👍

python/src/uagents/protocol.py
```diff
@@ -315,7 +315,7 @@ def manifest(self) -> Dict[str, Any]:
 
         for schema_digest, model in all_models.items():
             manifest["models"].append(
-                {"digest": schema_digest, "schema": model.model_json_schema()}
+                {"digest": schema_digest, "schema": model.schema()}
```
qati, line 318: why the change here? didn't we add model_json_schema so we don't have to change?

Archento: Apparently there was a missing positional argument meaning I forgot to add the class method decorator. It should work now without the change.

python/src/uagents/models.py
```diff
@@ -1,19 +1,22 @@
...
+
     @staticmethod
     def build_schema_digest(model: Union["Model", Type["Model"]]) -> str:
-        schema = model.model_json_schema()
-        digest = (
-            hashlib.sha256(json.dumps(schema, sort_keys=True).encode("utf8"))
-            .digest()
-            .hex()
-        )
+        schema = model.schema_json(indent=None, sort_keys=True)
```
qati, line 18: why change this? model_json_schema was added

Archento: Because that is a direct method call from the BaseModel and I don't want to rely on my mock methods.
Also we will need to refactor this method again in the near future when we tackle the manifest topic.

jrriehl (approved):
LGTM

python/pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "uagents"
-version = "0.13.1"
+version = "0.13.2"
```
jrriehl, line 3: You might want to revert this change so that you have something to commit for the release PR from branch `release/v0.13.2`.

Archento: You're right, thanks. I'll create a separate PR for that :)

python/src/uagents/__init__.py
```diff
@@ -1,4 +1,4 @@
 from .agent import Agent, Bureau  # noqa
 from .context import Context  # noqa
-from .models import Model  # noqa
+from .models import Model, Field  # noqa
```
lrahmani, line 3: Field v2 are not json serializable anymore
https://docs.pydantic.dev/latest/migration/#changes-to-pydanticfield

lrahmani: This most probabably implies that all of integrations and any other agents using Field would be broken with pydantic v2

python/src/uagents/models.py
```diff
@@ -1,19 +1,26 @@
...
 
 
+# reverting back to pydantic.v1 BaseModel for backwards compatibility
 class Model(BaseModel):
+    @classmethod
+    def model_json_schema(cls) -> str:
+        return cls.schema_json()
+
+    def model_dump_json(self) -> str:
+        return self.json()
```
lrahmani, line 14: Removed the `indent` and `sort_keys` to keep parity with v1
Example https://github.com/fetchai/uAgents/commit/9dc508d173dedc72ff3365692d4aafeb5f3e4de0#diff-93489236d31a4bf06237078577b1d223280256cf088366b24250fc36e62cf284L263

python/src/uagents/protocol.py
lrahmani, line 318: Restored the correct function as (I assume) it was mistakingly changed here
https://github.com/fetchai/uAgents/commit/110c1c921da040728d8aa5472b85ce29a6a532f4#diff-6043ffbea937f982ee0f219d8d51acc51c66a5d9ca9faef0a33bb26daf227f81L318
`schema` != `json_schema`
https://docs.pydantic.dev/1.10/usage/models/
https://docs.pydantic.dev/latest/migration/#changes-to-pydanticbasemodel

lrahmani (commented):
These changes should restore backward compatibility with uagents 0.12.2 for models, envelopes and protocols

python/src/uagents/__init__.py
Archento: Fields in v2 are different but should still be serialisable. What this tells me is that we have been using `Field` outside of its intended purpose.
By properly handling and adding metadata to the messages we should be able to avoid the use of Field altogether.
This can of cause happen down the line, it's just something to keep in mind.

lrahmani (approved):

lrahmani (approved):

PR #458, opened 2024-07-10
feat(core): 2nd part (registration) of feat/almanac-registration

Alejandro-Morales:


python/src/uagents/network.py
```diff
@@ -221,7 +221,7 @@ def get_expiry(self, address: str) -> int:
 
         return (expiry - height) * AVERAGE_BLOCK_INTERVAL
 
-    def get_endpoints(self, address: str) -> List[AgentEndpoint]:
+    def get_endpoints(self, address: str) -> Optional[List[AgentEndpoint]]:
```
jrriehl, line 224: Looks like this always returns a list.
```suggestion
    def get_endpoints(self, address: str) -> List[AgentEndpoint]:
```

python/src/uagents/registration.py
```diff
@@ -40,3 +61,162 @@ def _build_digest(self) -> bytes:
...
+        except Exception as e:
+            self._logger.error(f"Failed to register on Almanac API: {e}")
+
+        # schedule the ledger registration
+        try:
+            await self._ledger_policy.register(agent_address, protocols, endpoints)
+        except InsufficientFundsError:
+            self._logger.warning(
+                "Failed to register on Almanac contract due to insufficient funds"
+            )
```
jrriehl, line 219: We also need to reraise here in order to schedule the retry after a few seconds instead of the default registration interval.

jrriehl (approved):
LGTM

PR #465, opened 2024-07-18
feat(core): check supported almanac contract version against deployed…

Alejandro-Morales:


python/src/uagents/network.py
```diff
@@ -150,6 +150,17 @@ class AlmanacContract(LedgerContract):
...
+    def get_contract_version(self) -> str:
+        """
+        Get the version of the contract.
+
+        Returns:
+            str: The version of the contract.
+        """
+        query_msg = {"query_contract_state": {}}
+        response = self.query(query_msg)
+        return response["contract_version"]
```
Archento, line 162: I feel like this can lead to errors if the contract response is faulty and that we need to add a bit more checks in `self.query()` otherwise it looks good!

Archento (changes requested):
Looks good!
One small comment about code robustness but if that's added I'm good with the changes.

Archento (approved):
Much better now! Thanks

jrriehl (approved):

PR #490, opened 2024-08-07
feat(core): add message endpoint for inspector

Alejandro-Morales:
Received and sent local messages can now be retrieved through `http://localhost:<PORT>/messages` endpoint


Archento (changes requested):
This topic currently needs more discussion so we'll postpone the review for now.

python/src/uagents/agent.py
```diff
@@ -1147,7 +1151,10 @@ def __init__(
         self._port = port or 8000
         self._queries: Dict[str, asyncio.Future] = {}
         self._logger = get_logger("bureau", log_level)
-        self._server = ASGIServer(self._port, self._loop, self._queries, self._logger)
+        self._dispenser = Dispenser()
```
Archento, line 1154: Have you tested this with the bureau? I don't think this will work properly as each Agent already has a Dispenser which is used for sending messages. We need to trace where these messages will end up and potentially reroute the messages somehow to the bureaus instance.

python/src/uagents/asgi.py
```diff
@@ -149,6 +151,26 @@ async def handle_missing_content_type(self, headers: CaseInsensitiveDict, send):
                 }
             )
 
+    async def handle_get_messages(self, send):
+        """
+        Handle retrieval of stored messages.
+        """
+        messages = dispatcher.received_messages + self._dispenser.sent_messages
+        messages.sort(key=lambda x: x["timestamp"])
+        response = {"messages": [msg for msg in messages]}
```
Archento, line 160: I think we can safely send the whole messages list to the UI for the frontend to filter, e.g. based on a timespan.

python/src/uagents/communication.py
```diff
@@ -104,6 +105,17 @@ async def run(self):
                     sync=sync,
                 )
                 response_future.set_result(result)
+
+                self.sent_messages.append({
+                    "timestamp": time(),
```
Archento, line 110: Let's make that a pydantic `BaseModel` in envelope.py

```python

class EnvelopeHistoryEntry(BaseModel):
    timestamp: int = Field(default_factory=lambda: int(time.time()))
    version: int
    sender: str
    target: str
    session: UUID4
    schema_digest: str
    protocol_digest: Optional[str] = None
    payload: Optional[str] = None

    @field_serializer("session")
    def serialize_dt(self, session: UUID4, _info):
        return str(session)


class EnvelopeHistory(BaseModel):
    envelopes: List[EnvelopeHistoryEntry]
```

That way we can more easily create these:
```python
self.send_messages.append(EnvelopeHistoryEntry(**env))
```
 and when sending we just dump the whole model with:
```python
entry.model_dump_json()
```

python/src/uagents/dispatch.py
```diff
@@ -56,5 +58,15 @@ async def dispatch(
         for handler in self._sinks.get(destination, set()):
             await handler.handle_message(sender, schema_digest, message, session)
 
+        self.received_messages.append({
```
Archento, line 61: Same as above where `self.received_messages` would become `self.received_messages: EnvelopeHistory = []`

Archento (commented):
Nice! Good Progress :)
I've added some small proposals and we still need to discuss message retention but I think we're pretty much there.

Archento:
We can probably add a serialiser to the Envelope History List as well that automatically returns the sorted list when we call `.model_dump_json()`

```python
class EnvelopeHistory(BaseModel):
    envelopes: List[EnvelopeHistoryEntry]

    @field_serializer("envelopes", when_used="json")
    def serialize_envelopes_in_order(
        self, envelopes: List[EnvelopeHistoryEntry], _info
    ):
        return sorted(envelopes, key=lambda e: e.timestamp)
```

Archento (changes requested):
I'm inclined to move the storage of received messages back into the `asgi` module as it seems like it introduces unnecessary overhead by touching the variables twice when storing them.
How about we move the storing action to the asgi right in front of `await dispatcher.dispatch(...)`?
And since the envelope is not needed afterwards it would be ok to do the same approach as you did in the dispenser by using the env_dict and reuse that from there on.

dominic22:
- [ ] please also return the agent address from the `agent_info` endpoint

Archento (dismissed):
It looks good now and feels rounded 👍🏻

python/examples/19-rest-api/agent.py
```diff
@@ -9,7 +9,7 @@ class Request(Model):
 
 
 class Response(Model):
-    timestamp: datetime
+    timestamp: int
```
jrriehl, line 12:
```suggestion
    timestamp: float
```

python/src/uagents/agent.py
```diff
@@ -401,6 +406,23 @@ def __init__(
...
+            async def _handle_get_info(_ctx: Context):
+                return AgentInfo(
+                    agent_address=self.address,
+                    endpoints=self._endpoints,
+                    protocols=list(self.protocols.keys()),
+                )
+
+            @self.on_rest_get("/messages", EnvelopeHistory)  # type: ignore
+            async def _handle_get_messages(_ctx: Context):
+                return self._message_cache
```
jrriehl, line 422: Nice, exactly how I was hoping this would look!

python/src/uagents/envelope.py
```diff
@@ -110,3 +117,49 @@ def _digest(self) -> bytes:
...
+    def apply_retention_policy(self):
+        """Remove entries older than 24 hours"""
+        cutoff_time = time.time() - 86400
+        self.envelopes = [e for e in self.envelopes if e.timestamp > cutoff_time]
+
+    @field_serializer("envelopes", when_used="always")
+    def serialize_envelopes_in_order(
+        self, envelopes: List[EnvelopeHistoryEntry], _info
+    ):
+        return sorted(envelopes, key=lambda e: e.timestamp)
```
jrriehl, line 165: I'm wondering if we could manage both the retention policy and the serialization somewhat more efficiently by storing the envelopes as a `dict` with timestamp as the key?

jrriehl (commented):
Really nice! Just one question / suggestion about the data type for storing the history.

python/src/uagents/envelope.py
Archento: Good point. I'd argue that it doesn't matter much if we use models instead of dictionaries but I see a problem with the loop in line 159 now since that can get inefficient with a large number of messages within the retention period.
Since we decided against a maximum number of messages it's hard to identify an upper ceiling so some form of iteration over the dataset will be necessary.

We'll need a Model for validation in the Rest endpoint anyway so it made sense to do the sorting implicitly here. (although in my tests sorting isn't actually needed as the messages will be added to the list in the order they appear...)

jrriehl: Yes, that line 159 was mostly what I had in mind. With the envelopes sorted by timestamp, we could start from the oldest and stop at the first envelope whose timestamp puts it inside the retention window.

I think that line 165 could then become something like `return list(envelopes.values())`.

Archento: refactored the retention mechanism and dropped the serialiser method as that was not needed any more.

Archento (approved):

PR #507, opened 2024-08-16
chore(core): verify sync envelope if signed

jrriehl:
## Proposed Changes

A previous update left open the possibility that an envelope sent as a sync response with a signature would not be verified. With this change, we ensure that any envelope that contains a signature gets verified.

## Types of changes

- [ ] Documentation update.

## Checklist

### DIFF
--- a/python/src/uagents/communication.py
+++ b/python/src/uagents/communication.py
@@ -163,10 +163,11 @@ async def send_exchange_envelope(
                     success = resp.status == 200
                     if success:
                         if sync:
-                            # If the message is synchronous but not verified, return the envelope
                             env = Envelope.model_validate(await resp.json())
-                            if env.signature is None:
-                                return env
+                            if env.signature and not env.verify():
+                                errors.append(
+                                    "Received response envelope that failed verification"
+                                )
                             return await dispatch_sync_response_envelope(env)
                         return MsgStatus(
                             status=DeliveryStatus.DELIVERED,
--- a/python/src/uagents/resolver.py
+++ b/python/src/uagents/resolver.py
@@ -281,9 +281,9 @@ class AlmanacApiResolver(Resolver):
 
             if response.status_code != 200:
                 if response.status_code != 404:
-                    LOGGER.warning(
-                        f"Failed to resolve agent {address} from {self._almanac_api_url},"
-                        f"querying Almanac contract..."
+                    LOGGER.debug(
+                        f"Failed to resolve agent {address} from {self._almanac_api_url}, "
+                        f"resolvong via Almanac contract..."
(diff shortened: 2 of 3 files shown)


python/src/uagents/communication.py
```diff
@@ -163,10 +163,12 @@ async def send_exchange_envelope(
                     success = resp.status == 200
                     if success:
                         if sync:
-                            # If the message is synchronous but not verified, return the envelope
                             env = Envelope.model_validate(await resp.json())
-                            if env.signature is None:
-                                return env
```
jrriehl, line 169: This return is redundant because it should only apply when the message is sent from somethere other than an agent (i.e. the `query` function). In this case, the dispatcher should have no sinks and the function `dispatch_sync_response_envelope` will return the envelope.

python/src/uagents/communication.py
```diff
@@ -163,10 +163,12 @@ async def send_exchange_envelope(
...
+                            if env.signature and not env.verify():
+                                errors.append(
+                                    "Received response envelope that failed verification"
+                                )
+                                continue
```
jrriehl, line 171: This is considered to be a failed delivery even though a response was received and is consistent with the case when an unsigned envelope is received from an agent via the asgi server: https://github.com/fetchai/uAgents/blob/main/python/src/uagents/asgi.py#L273-L289.

python/src/uagents/resolver.py
```diff
@@ -281,9 +281,9 @@ async def _api_resolve(self, destination: str) -> Tuple[Optional[str], List[str]
 
             if response.status_code != 200:
                 if response.status_code != 404:
-                    LOGGER.warning(
-                        f"Failed to resolve agent {address} from {self._almanac_api_url},"
-                        f"querying Almanac contract..."
+                    LOGGER.debug(
```
jrriehl, line 284: As long as the agent endpoint gets resolved somehow, I think this shouldn't rise to the WARNING level.

python/src/uagents/communication.py
```diff
@@ -163,10 +163,12 @@ async def send_exchange_envelope(
                     success = resp.status == 200
                     if success:
                         if sync:
-                            # If the message is synchronous but not verified, return the envelope
                             env = Envelope.model_validate(await resp.json())
-                            if env.signature is None:
-                                return env
+                            if env.signature and not env.verify():
```
Archento, line 167: If verification fails (`env.verify()`) it'll raise an Exception (ValueError, BadSignatureError or BadDigestError) so we need to catch that as the only bool ever to be returned is `True`

Archento (approved):

PR #516, opened 2024-08-28
fix(core): asgi error message due to missing body

Archento:
A new bug was introduced in the last PR where you'll receive an error log message from the ASGI server due to an incomplete message.

`ERROR:    ASGI callable returned without completing response.`

These logs would be triggered whenever a message has been sent as the callee would send an incomplete acknowledgement.


Alejandro-Morales (approved):
Looks good and works fine!

PR #531, opened 2024-09-17
fix(core): consistently generated session ids

ejfitzgerald:
### Summary of the Problem

Internally, the uAgents have a concept of an Internal and external context variants. The internal one was created and cached inside the agent object. 

Due to the nature of the cached internal context, this would leads to some interesting outcomes specifically for the session id. In some cases (like a startup event) the session id can be `None`. Additionally, some session ids values are cached depending on the order of how handlers are called.

All of the above means that being able to reliable use the session id for chaining is difficult.

### Consistent Generation of Session IDs

Semantically, this change ensures that the following rule is always applied: 

**_The session id is always taken from the triggering envelope, when this is not possible a new session id is always generated_**

As a result of this, the code is updated so that the context is always built before it is used. This was always the case for the external context, but now is true for the internal context as well. This consistency eliminates some of the quirky cached behaviour that we had beforehand.

More importantly, however, this ensures that the session id can be used by the agent code to track responses from chained sessions.

## Linked Issues

n/a

## Types of changes

- [ ] Documentation update.

## Checklist

### If applicable

- [ ] I have added/updated the documentation

## Further comments


XaviPeiro (approved):
Nothing necessary to change in here, quite clean and the change will allow us to apply specific logic on every internal use-case.

A general advise is use and enforce name parameters, it avoids errors when refactoring and increases intelligibility quite much.

python/src/uagents/agent.py
```diff
@@ -1163,7 +1177,7 @@ async def _process_message_queue(self):
             )
 
             context = ExternalContext(
-                agent=self._ctx.agent,
+                agent=self,
```
Archento, line 1180: This needs to be the agent representation not the agent itself

ejfitzgerald:
@Archento I have modified the code so that it is not coupled with the `Agent` interface. When you are back on Monday can we aim to review and land this please 🙏

Archento (approved):

jrriehl (approved):
LGTM

PR #549, opened 2024-10-10
feat(core): add agent metadata property

Archento:
## Proposed Changes

Add metadata property to agent constructor.
The initial version will support metadata in the following form:
```python
class AgentGeolocation(BaseModel):
    latitude: float
    longitude: float
    radius: int

class AgentMetadata(BaseModel):
    geolocation: AgentGeolocation
```

Other metadata will be ignored which means the following agent will produce the respective output:
```python
agent = Agent(
    metadata={
        "geolocation": {
            "latitude": 52.507410,
            "longitude": 13.378240,
            "radius": 50,
        },
        "mobility": True,
        "useless_information": "This is a test",
    },
)
print(agent.metadata)
>>> {'geolocation': {'latitude': 52.50741, 'longitude': 13.37824, 'radius': 50}}
```

## Types of changes

- [ ] Documentation update.

## Checklist

### If applicable

- [x] I have checked that code coverage does not decrease

### DIFF
--- a/python/src/uagents/agent.py
+++ b/python/src/uagents/agent.py
@@ -21,7 +21,7 @@ import requests
 from cosmpy.aerial.client import LedgerClient
 from cosmpy.aerial.wallet import LocalWallet, PrivateKey
 from cosmpy.crypto.address import Address
-from pydantic import BaseModel, ValidationError
+from pydantic import ValidationError
 
 from uagents.asgi import ASGIServer
 from uagents.communication import Dispenser
@@ -57,6 +57,7 @@ from uagents.storage import KeyValueStore, get_or_create_private_keys
 from uagents.types import (
     AgentEndpoint,
     AgentInfo,
+    AgentMetadata,
     EventCallback,
     IntervalCallback,
     JsonStr,
@@ -513,7 +514,9 @@ class Agent(Sink):
         else:
             self._wallet_messaging_client = None
 
-    def _initialize_metadata(self, metadata: Optional[Dict[str, Any]]):
+    def _initialize_metadata(
+        self, metadata: Optional[Dict[str, Any]]
+    ) -> Dict[str, Any]:
         """
         Initialize the metadata for the agent.
 
@@ -521,18 +524,13 @@ class Agent(Sink):
         model ensures that the metadata is valid and complete.
 
         Args:
-            metadata (Dict[str, Any]): The metadata to include in the agent object.
+            metadata (Optional[Dict[str, Any]]): The metadata to include in the agent object.
+
+        Returns:
+            Dict[str, Any]: The filtered metadata.
         """
         if not metadata or "geolocation" not in metadata:
             return {}
-
-        class AgentGeolocation(BaseModel):
-            latitude: float
-            longitude: float
-            radius: int
-
-        class AgentMetadata(BaseModel):
-            geolocation: AgentGeolocation
 
         try:
             model = AgentMetadata.model_validate(metadata, strict=True)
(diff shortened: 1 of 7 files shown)


python/src/uagents/agent.py
```diff
@@ -508,6 +513,36 @@ def initialize_wallet_messaging(
...
+        The metadata is filtered to include only location-based metadata and the
+        model ensures that the metadata is valid and complete.
+
+        Args:
+            metadata (Dict[str, Any]): The metadata to include in the agent object.
+        """
+        if not metadata or "geolocation" not in metadata:
+            return {}
+
+        class AgentGeolocation(BaseModel):
```
Dacksus, line 529: We will definitely use this set (lat, lng, radius) more often at different places in the Agent. I like the idea of making it available in the `types.py`

python/src/uagents/agent.py
```diff
@@ -508,6 +513,36 @@ def initialize_wallet_messaging(
         else:
             self._wallet_messaging_client = None
 
+    def _initialize_metadata(self, metadata: Optional[Dict[str, Any]]):
```
jrriehl, line 516:
```suggestion
    def _initialize_metadata(self, metadata: Optional[Dict[str, Any]]) -> Dict[str, Any]
```

python/src/uagents/agent.py
```diff
@@ -508,6 +513,36 @@ def initialize_wallet_messaging(
...
+        """
+        Initialize the metadata for the agent.
+
+        The metadata is filtered to include only location-based metadata and the
+        model ensures that the metadata is valid and complete.
+
+        Args:
+            metadata (Dict[str, Any]): The metadata to include in the agent object.
+        """
```
jrriehl, line 525:
```suggestion
        Args:
            metadata (Dict[str, Any]): The metadata to include in the agent object.
        Returns:
            Dict[str, Any]: The filtered metadata.
        """
```

python/src/uagents/agent.py
Archento: done

python/src/uagents/types.py
```diff
@@ -55,6 +56,30 @@ class RestHandlerDetails(BaseModel):
...
+        float,
+        Field(strict=True, ge=-90, le=90, allow_inf_nan=False),
+    ]
+    longitude: Annotated[
+        float,
+        Field(strict=True, ge=-180, le=180, allow_inf_nan=False),
+    ]
+    radius: Annotated[
+        float,
+        Field(strict=True, ge=0, allow_inf_nan=False),
```
jrriehl, line 70: Might there be a case where infinite radius makes sense? Something entity that is available everywhere?

python/src/uagents/types.py
```diff
@@ -55,6 +56,30 @@ class RestHandlerDetails(BaseModel):
...
+    ] = 0
+
+    @field_serializer("latitude", "longitude")
+    def serialize_precision(self, val: float) -> float:
+        """Round the latitude and longitude to 6 decimal places."""
+        return float(f"{val:.6f}")
```
jrriehl, line 76: What's the reason for this?

jrriehl (commented):
Looks good overall! Just a minor comment and question.

python/src/uagents/types.py
Archento: I'd be restrictive for now and open this up once we see a use-case for it.
Right now there will potentially only be local scenarios and depending on the architecture we might introduce clusters that cover broader ranges. We can get back to this at the latest when we introduce polygonal zones :)

python/src/uagents/types.py
Archento: A precision of 6 digits after the comma equates to 10cm so I think that should be enough.
There is a funny [comparison](https://xkcd.com/2170/) from xkcd for that 😄

jrriehl (approved):

PR #551, opened 2024-10-11
feat(core): add batch almanac api and contract registrations for Bureau

jrriehl:
## Proposed Changes

When running the Bureau, the Bureau will now use the Almanac API's batch registration endpoint by default to register all agents in a single request. Additionally, if a wallet is provided in the Bureau's init args, it will also send the Almanac contract registration transactions in a batch. By default, with no wallet address provided, the agents will register themselves as before, so this update should be fully backwards compatible.

## Types of changes

- [ ] Documentation update.

## Checklist

### If applicable

- [ ] I have added/updated the documentation


python/src/uagents/registration.py
```diff
@@ -133,6 +141,56 @@ async def register(
...
 
+class BatchAlmanacApiRegistrationPolicy(AgentRegistrationPolicy):
+    def __init__(
+        self, almanac_api: Optional[str] = None, logger: Optional[logging.Logger] = None
+    ):
+        self._almanac_api = almanac_api or ALMANAC_API_URL
+        self._attestations: List[AgentRegistrationAttestation] = []
+        self._logger = logger or logging.getLogger(__name__)
+
+    def add_agent(self, agent: Any):
```
jrriehl, line 152: I hesitated to pass the agent here, and may still reconsider, but I wanted to reuse this interface for the BatchLedgerRegistrationPolicy, which may require other components of the agent like wallet address.

Archento: I'd also avoid passing the agent object since the Bureau should have access to all the information of the agent already.
And looking at the other ABC I'm almost certain that we can reuse the interface as that already covers the Almanac API as well as the Contract.

The signing can be done in the `agent.py` so that we don't need to pass the identity here OR we can do the same as in the `Envelope` and just forward the signing Callback.

python/src/uagents/registration.py
```diff
@@ -133,6 +141,56 @@ async def register(
...
+        attestation.sign(agent._identity)
+        self._attestations.append(attestation)
+
+    async def register(self):
+        if not self._attestations:
+            return
+
+        attestations = [a.model_dump() for a in self._attestations]
+
+        async with aiohttp.ClientSession() as session:
```
Archento, line 168: This to me screams like it wants to be a separate function where you give it an endpoint and a list of attestations (which can have one entry for single registration and multiple for the batch)
But up to you

Archento (commented):
Love it!

I just have two minor comments/discussion points but other than that I'm good with the feature 👍🏻

python/src/uagents/registration.py
jrriehl: Done, reused `almanac_api_post`.

python/src/uagents/agent.py
```diff
@@ -1362,31 +1407,114 @@ def __init__(
...
+            ):
+                raise ValueError(
+                    "Argument 'wallet' must be provided when using "
+                    "the batch ledger registration policy."
+                )
+            self._registration_policy = registration_policy
+        else:
+            self._registration_policy = DefaultBatchRegistrationPolicy(
+                ledger or get_ledger(test),
+                wallet,
```
Archento, line 1428: just for confirmation: if you don't supply a wallet it'll only use the API and not the Ledger.

python/src/uagents/registration.py
```diff
@@ -167,6 +188,41 @@ async def register(
...
+        self._almanac_api = almanac_api or ALMANAC_API_URL
+        self._attestations: List[AgentRegistrationAttestation] = []
+        self._logger = logger or logging.getLogger(__name__)
+
+    def add_agent(self, agent_info: AgentInfo, identity: Identity):
+        attestation = AgentRegistrationAttestation(
+            agent_address=agent_info.agent_address,
+            protocols=list(agent_info.protocols),
+            endpoints=agent_info.endpoints,
+            metadata=coerce_metadata_to_str(agent_info.metadata),
```
Archento, line 208: This should (for now) only include geolocation. See `AlmanacApiRegistrationPolicy.register()`

python/src/uagents/registration.py
```diff
@@ -253,18 +330,76 @@ def _sign_registration(self, current_time: int) -> str:
...
+        self._logger = logger or logging.getLogger(__name__)
+        self._records: List[AlmanacContractRecord] = []
+        self._identities: Dict[str, Identity] = {}
+
+    def add_agent(self, agent_info: AgentInfo, identity: Identity):
+        agent_record = AlmanacContractRecord(
+            agent_address=agent_info.agent_address,
+            protocols=agent_info.protocols,
+            endpoints=agent_info.endpoints,
+            metadata=agent_info.metadata,
```
Archento, line 360: same as above

Archento (changes requested):
Looks great!
Also a lot of changes have been made so I hope to have gotten every detail.
The only thing that stood out for me is the handling of metadata as I think that has not been decided yet.

python/src/uagents/agent.py
jrriehl: Not quite, the agents will register themselves on the ledger as they did before this update.

Archento (approved):
Great work!

PR #561, opened 2024-10-16
feat(core): enable update registration policy after agent init

Archento:
## Proposed Changes

This addresses an issue where you'll want to set the registration policy manually but were not able (or rather it was difficult) do so because of the missing agent information that you'll only get after initialising an agent.

It also fixes an issue with the `on_rest_post` decorator where the `Request` Model was erroneously set as Optional.

## Types of changes

- [ ] Documentation update.

## Checklist

### If applicable

### DIFF
--- a/python/.pre-commit-config.yaml
+++ b/python/.pre-commit-config.yaml
@@ -1,7 +1,6 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.6.0
-    files: ^python/
+    rev: v0.6.9
     hooks:
       - id: ruff
         args: [--fix]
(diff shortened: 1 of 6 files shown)


python/.pre-commit-config.yaml
```diff
@@ -1,7 +1,6 @@
 repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.6.0
-    files: ^python/
```
Archento, line 4: I think we can exclude this restriction now since integrations have been or will be moved.

jrriehl (approved):
LGTM

PR #564, opened 2024-10-21
feat(core): add is active almanac update

lrahmani:


python/src/uagents/registration.py
```diff
@@ -293,3 +293,42 @@ async def register(
...
+            json.dumps(
+                self.model_dump(exclude={"signature"}),
+                sort_keys=True,
+                separators=(",", ":"),
+            ).encode("utf-8")
+        )
+        return sha256.digest()
+
+
+async def update_agent_status(update: AgentStatusUpdate, almanac_api: str):
```
lrahmani, line 326: With some refactoring this could be integrated into the `AgentRegistrationPolicy`

python/src/uagents/registration.py
```diff
@@ -293,3 +293,42 @@ async def register(
...
+    async with aiohttp.ClientSession() as session:  # noqa: SIM117
+        async with session.post(
```
lrahmani, line 328: It may be worth adding retries here

python/src/uagents/registration.py
```diff
@@ -293,3 +293,41 @@ async def register(
...
+            json.dumps(
+                self.model_dump(exclude={"signature"}),
+                sort_keys=True,
+                separators=(",", ":"),
+            ).encode("utf-8")
+        )
+        return sha256.digest()
+
+
+async def update_agent_status(status: AgentStatusUpdate, almanac_api: str):
```
lrahmani, line 325: With some refactoring this could be integrated into the `AgentRegistrationPolicy`

Archento: I think this needs to be separate as the status update and attestation is only relevant for the Almanac API but the `AgentRegistrationPolicy` covers all registration aspects.

python/src/uagents/registration.py
Archento: It depends on how important this information is. We need to accept the fact that that there will always be situations in which an agent cannot send an update to the Almanac API but I am still for adding retries here. I'd say 3 should be enough with the exponential backoff.

python/src/uagents/registration.py
```diff
@@ -293,3 +293,40 @@ async def register(
         except Exception as e:
             self._logger.error(f"Failed to register on Almanac contract: {e}")
             raise
+
+
+class AgentStatusUpdate(BaseModel):
```
Archento, line 298: This is the same structure as the `AgentRegistrationAttestation` so we should make both inherit from the same model.

Archento (changes requested):
Looks good so far.

We should refactor the registration code a bit to remove the duplicate code bits but once that's done it's good!

Archento (approved):

python/src/uagents/registration.py
```diff
@@ -49,6 +95,29 @@ def coerce_metadata_to_str(
     return out
 
 
+async def use_almanac_api(
```
lrahmani, line 98: Maybe this can be renamed to indicate that it is sending a post message, `almanac_api_post` for e.g.

python/src/uagents/registration.py
```diff
@@ -23,6 +24,51 @@
...
+        )
+        return sha256.digest()
+
+
+class AgentRegistrationAttestation(VerifiableModel):
+    protocols: List[str]
+    endpoints: List[AgentEndpoint]
+    metadata: Optional[Dict[str, Union[str, Dict[str, str]]]] = None
+
+    @field_serializer("protocols")
```
lrahmani, line 59: Is there a way to keep the original object unchanged and only apply these to a copy with will be used for generating the digest?

Archento: I can think of a couple but do we need to keep these separated? The order of the data has no impact on the following processing as far as I’m aware and this method provides some consistency as every time the model is dumped it produces the same result.

jrriehl: Okay, at first I was thinking that there could be some functional reason for ordering the endpoints, but probably that only applies to the agent itself and not the attestation, so I think we can keep these field serializers.

Archento: I created these serializers because in the original digest function the protocols and endpoints were ordered. Looking at it now it makes no sense they exist in the first place.

The connotation of the digest is inherently different from that of a manifest where we give leeway with the order of the fields etc. by always ordering before calculating the digest. Here, it's just a means of creating a verification method for integrity. Also, in the endpoint we actually remove all related data first before taking it 1:1 from the attestation so there is no ordering necessary here as well.

For that reason I'll remove them.

Archento (approved):

PR #582, opened 2024-11-22
chore(example): remove obsolete fund_agents_if_low

Dacksus:
funds are added automatically on registration, no need to explicitly do this anymore.


Archento:
I think this is the only example where it actually makes sense to have the former (now redundant) function in. This is because we'll need to fund a separate wallet and not the agents' wallet itself during startup (to make sure the example works).
If there are no objections we can close this issue.

Archento:
Closing this now

PR #588, opened 2024-12-04
feat(core): updates to mailbox and agentverse registration process

jrriehl:
## Proposed Changes

- Add updates for new mailbox connection workflow
- Add support for proxy and custom agent registration
- Remove websocket connection, which hasn't been supported for a long time anyway
- Delete mailbox messages immediately after handling them instead of processing them in a queue
- Handle CORS preflight request

## Linked Issues

## Types of changes

- [ ] Documentation update.

## Checklist

### If applicable

### DIFF
--- a/python/examples/11-mailbox-agents/bob.py
+++ b/python/examples/11-mailbox-agents/bob.py
@@ -11,18 +11,12 @@ ALICE_ADDRESS = "paste_alice_address_here"
 # Generate a second seed phrase (e.g. https://pypi.org/project/mnemonic/)
 SEED_PHRASE = "put_your_seed_phrase_here"
 
-# Copy the address shown below
-print(f"Your agent's address is: {Agent(seed=SEED_PHRASE).address}")
-
-# Then go to https://agentverse.ai, register your agent in the Mailroom
-# and copy the agent's mailbox key
-AGENT_MAILBOX_KEY = "put_your_AGENT_MAILBOX_KEY_here"
-
 # Now your agent is ready to join the agentverse!
 agent = Agent(
     name="bob",
     seed=SEED_PHRASE,
-    mailbox=f"{AGENT_MAILBOX_KEY}@https://agentverse.ai",
+    port=8001,
+    endpoint="http://agentverse.ai/v1/submit",
 )
 
 
--- a/python/examples/11-mailbox-agents/bob.py
+++ b/python/examples/11-mailbox-agents/bob.py
@@ -8,16 +8,9 @@ class Message(Model):
 # Copy ALICE_ADDRESS generated in alice.py
 ALICE_ADDRESS = "paste_alice_address_here"
 
-# Generate a second seed phrase (e.g. https://pypi.org/project/mnemonic/)
-SEED_PHRASE = "put_your_seed_phrase_here"
 
 # Now your agent is ready to join the agentverse!
-agent = Agent(
-    name="bob",
-    seed=SEED_PHRASE,
-    port=8001,
-    endpoint="http://agentverse.ai/v1/submit",
-)
+agent = Agent(name="bob", port=8009, mailbox=True)
 
 
 @agent.on_interval(period=2.0)
--- a/python/src/uagents/agent.py
+++ b/python/src/uagents/agent.py
@@ -424,6 +424,7 @@ class Agent(Sink):
 
             @self.on_rest_post("/prove", AgentverseConnectRequest, RegistrationResponse)  # type: ignore
             async def _handle_prove(_ctx: Context, token: AgentverseConnectRequest):
+                print(self._agentverse)
                 return await register_in_agentverse(
                     token, self._identity, self._endpoints, self._agentverse
                 )
--- a/python/src/uagents/agent.py
+++ b/python/src/uagents/agent.py
@@ -44,6 +44,7 @@ from uagents.mailbox import (
     AgentverseConnectRequest,
     MailboxClient,
     RegistrationResponse,
+    is_mailbox_agent,
     register_in_agentverse,
 )
 from uagents.models import ErrorMessage, Model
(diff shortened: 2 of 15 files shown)


python/src/uagents/agent.py
```diff
@@ -416,6 +422,12 @@ async def _handle_get_info(_ctx: Context):
             async def _handle_get_messages(_ctx: Context):
                 return self._message_cache
 
+            @self.on_rest_post("/prove", AgentverseUserToken, RegistrationResponse)  # type: ignore
```
Dacksus: this will be triggered by the inspector?

python/src/uagents/config.py
```diff
@@ -43,6 +45,21 @@
 DEFAULT_SEARCH_LIMIT = 100
 
 
+AgentType = Literal["hosted", "local", "mailbox", "proxy", "custom"]
```
Dacksus, line 48: "local" means not connection with the agentverse whatsoever and "custom" only that it's bound to an agentverse account?

python/src/uagents/mailbox.py
```diff
@@ -1,101 +1,171 @@
...
+    user_token: AgentverseUserToken,
+    identity: Identity,
+    endpoints: list[AgentEndpoint],
+    agentverse: AgentverseConfig,
+) -> Optional[RegistrationResponse]:
+    """
+    Registers agent in Agentverse
+    """
+    async with aiohttp.ClientSession() as session:
+        # get challenge
```
Dacksus, line 65: Why do we need this challenge? We want to prove that the caller of the registration endpoint is in fact the agent owning the private key to the given address? 
Wouldn't it be sufficient to simply sign the RegistrationRequest below instead of the challenge to save one round trip? We would still be able to verify if the signature matches the given address. A replay attack would only make sense for an adversary if either address or endpoint could be altered, which they can't because they would be signed.

python/src/uagents/agent.py
Dacksus: And either we display the token in the inspector, request the user to update the code and to restart the agent, or would need to write the key to a file and update it in memory?

jrriehl: Yes, the inspector will use this endpoint and provide a code snippet for how to add the access token to the agent.

python/src/uagents/config.py
jrriehl: Yes, exactly. There is probably some room for improvement in these names.

python/src/uagents/mailbox.py
```diff
@@ -1,101 +1,171 @@
...
+    uuid: UUID4
+    envelope: Envelope
+    received_at: datetime
+    expires_at: datetime
 
-        Returns: The http prefix of the mailbox server {http, https}.
-        """
-        return self._agent.mailbox["http_prefix"]
+
+async def register_in_agentverse(
```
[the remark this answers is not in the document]
jrriehl: We require the agent type as an input parameter, which will be provided by the inspector.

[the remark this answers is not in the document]
Dacksus: Where do we require it? Should it be part of the RegistrationRequest?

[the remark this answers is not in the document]
Dacksus: nvm

python/src/uagents/mailbox.py
```diff
@@ -1,101 +1,234 @@
...
+                logger.info(
+                    f"Successfully registered as {request.agent_type} agent in Agentverse"
+                )
+
+    if request.agent_type == "mailbox" and not is_mailbox_agent(endpoints, agentverse):
+        logger.exception(
+            f"Agent endpoints {endpoints} do not match registered agent type: {request.agent_type}"
+            f"Please restart agent with endpoint='{agentverse.url}/v1/submit'"
+        )
+    elif request.agent_type == "proxy" and not is_proxy_agent(endpoints, agentverse):
```
Dacksus: Don't we expect proxy agents to provide their own reachable endpoints (i.e., not containing the agentverse url), so they can be reached after the redirect from the mailbox?

python/src/uagents/mailbox.py
jrriehl: It's possible that another approach could also work here, but I'd suggest we keep this one for now and maybe revisit later on whether a simpler method can provide sufficient security.

python/examples/11-mailbox-agents/bob.py
```diff
@@ -11,18 +11,12 @@ class Message(Model):
...
 SEED_PHRASE = "put_your_seed_phrase_here"
 
-# Copy the address shown below
-print(f"Your agent's address is: {Agent(seed=SEED_PHRASE).address}")
-
-# Then go to https://agentverse.ai, register your agent in the Mailroom
-# and copy the agent's mailbox key
-AGENT_MAILBOX_KEY = "put_your_AGENT_MAILBOX_KEY_here"
-
 # Now your agent is ready to join the agentverse!
```
Dacksus, line 14: Since we're not configuring anything that's called "mailbox" it is less obvious to see what kind of agent is being configured, if you don't already know. Also if I understand correctly, the only way now we're officially supporting is to use the local inspector (?). So I'd suggest to add more commentary too guide the user:

\# By specifying the agentverse submit endpoint, your agent will automatically know it's a mailbox agent. In order to complete registration, you need to visit your agents local inspector interface and connect it with your agentverse account.

We should add or link to more documentation regarding this process as soon as it is available.

Dacksus (approved):

python/tests/test_config.py
```diff
@@ -1,20 +1,36 @@
...
+    Agent(proxy=True, endpoint="http://some_url"),
+    Agent(mailbox=True, proxy=True),
+]
+
+expected_endpoints = [
+    [],
+    [AgentEndpoint(url=MAILBOX_ENDPOINT, weight=1)],
+    [AgentEndpoint(url="http://some_url", weight=1)],
+    [],
+    [],
```
Dacksus, line 28: Not AgentEndpoint(url="https://staging.agentverse.ai", weight=1)?

Dacksus (approved):
Minor comment regarding the tests, otherwise LGTM! Creating a mailbox/proxy agent now feels much cleaner with the boolean options and the AgentverseConfig

python/tests/test_config.py
jrriehl: No, because agentverse configuration is not completely separate from endpoint configuration, which I think is an improvement.

Dacksus: I assume you meant "not" = "now

python/src/uagents/config.py
```diff
@@ -43,15 +47,53 @@
...
+
+
+class AgentverseConfig(BaseModel):
```
Dacksus, line 53: Do we actually still need to distinguish between `protocol` and `http_prefix` if we drop support for websockets? To maintain backwards compatibility, we could keep protocol, but make it at least Optional

jrriehl: No, we don't need that any more. I was thinking it's not bad to keep it around in case we bring back websockets, but mostly just didn't get around to removing it.

Dacksus: realized that it doesn't matter for a user, fine with leaving it for the time being

python/tests/test_config.py
jrriehl: Yes, sorry. :smile:

PR #595, opened 2024-12-19
fix(core): proxy endpoint flow

jrriehl:
## Proposed Changes

Endpoint needs to be supplied to `/prove` endpoint for proxy agent.

## Types of changes

- [ ] Documentation update.

## Checklist

### If applicable

### DIFF
--- a/python/src/uagents/mailbox.py
+++ b/python/src/uagents/mailbox.py
@@ -24,6 +24,7 @@ AgentType = Literal["mailbox", "proxy", "custom"]
 class AgentverseConnectRequest(Model):
     user_token: str
     agent_type: AgentType
+    endpoint: Optional[str] = None
 
 
 class ChallengeRequest(BaseModel):
@@ -50,7 +51,7 @@ class RegistrationRequest(BaseModel):
     challenge: str
     challenge_response: str
     agent_type: AgentType
-    endpoints: Optional[list[AgentEndpoint]] = None
+    endpoint: Optional[str] = None
 
 
 class RegistrationResponse(Model):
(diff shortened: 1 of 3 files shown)


python/src/uagents/mailbox.py
```diff
@@ -24,6 +24,7 @@
 class AgentverseConnectRequest(Model):
     user_token: str
     agent_type: AgentType
+    endpoint: Optional[str] = None
```
jrriehl, line 27: Changed this to a string to simplify things, because it gets to complicated to next a pydantic v2 model inside of a pydantic v1 model.

python/src/uagents/mailbox.py
```diff
@@ -164,12 +152,6 @@ async def register_in_agentverse(
             request.user_token, identity.address, agent_details, agentverse
         )
 
-    if request.agent_type == "mailbox" and not is_mailbox_agent(endpoints, agentverse):
```
jrriehl, line 167: We no longer have the agent's endpoints, so it doesn't make sense to do this check here anymore.

python/src/uagents/mailbox.py
```diff
@@ -83,22 +84,9 @@ def is_mailbox_agent(
     return any([f"{agentverse.url}/v1/submit" in ep.url for ep in endpoints])
 
 
-def is_proxy_agent(
```
jrriehl, line 86: Unused

python/src/uagents/mailbox.py
Dacksus: Can it only be a single endpoint or several?

Dacksus (approved):
LGTM, just one comment

python/src/uagents/mailbox.py
jrriehl: Just a single endpoint now.

PR #597, opened 2024-12-23
feat(core): Introducing uagents-core

qati:
## Proposed Changes

_[describe the changes here...]_

## Linked Issues

## Types of changes

- [ ] Documentation update.

## Checklist

### If applicable

## Further comments


libs/core/uagents_core/config.py
```diff
@@ -0,0 +1,6 @@
+
+
+DEFAULT_AGENTVERSE_URL = "https://agentverse.ai"
+DEFAULT_ALMANAC_API_PATH = "/v1/almanac"
+DEFAULT_MAILBOX_API_PATH = "/v1/agents"
```
jrriehl, line 5: Mailbox API includes several other paths as well. Maybe something like this would be more specific?
```suggestion
DEFAULT_REGISTRATION_PATH = "/v1/agents"
```

libs/core/uagents_core/logger.py
jrriehl, line 1: I guess this is only used in utils, so maybe should be moved there?

libs/core/uagents_core/utils/registration.py
```diff
@@ -0,0 +1,229 @@
...
+from uagents_core.logger import get_logger
+
+
+
+
+logger = get_logger("uagents_core.utils.registration")
+
+
+class VerifiableModel(BaseModel):
+    agent_address: str
```
jrriehl, line 27: This will likely need an update very soon based on this PR: https://github.com/fetchai/uAgents/pull/603.

libs/core/uagents_core/utils/registration.py
```diff
@@ -0,0 +1,229 @@
...
+    identity: Identity,
+    url: str,
+    agentverse_token: str,
+    agent_title: str,
+    readme: str,
+    *,
+    protocol_digest: str,
+    agentverse_url: Optional[str] = None,
+    agent_type: AgentType = "custom",
+):
```
jrriehl, line 96: Can we at least use the same API that we use in the async version of this: https://github.com/fetchai/uAgents/blob/main/python/src/uagents/mailbox.py#L87-L92?

libs/core/uagents_core/utils/communication.py
```diff
@@ -0,0 +1,87 @@
...
+def send_message_to_agent(
+    sender: Identity,
+    target: str,
+    payload: Any,
+    protocol_digest: str,
+    model_digest: str,
+    session: UUID = uuid4(),
+    *,
+    agentverse_url: Optional[str] = None,
+):
```
jrriehl, line 48: I don't mind having a duplicate sync version of this function, but I'd suggest we use the same or at least very similar naming conventions and API as the async version: https://github.com/fetchai/uAgents/blob/main/python/src/uagents/communication.py#L267-L275

libs/core/uagents_core/utils/communication.py
```diff
@@ -0,0 +1,87 @@
...
+    """
+    Send a message to an agent.
+    :param session: The unique identifier for the dialogue between two agents
+    :param sender: The identity of the sender.
+    :param target: The address of the target agent.
+    :param protocol_digest: The digest of the protocol that is being used
+    :param model_digest: The digest of the model that is being used
+    :param payload: The payload of the message.
+    :return:
+    """
```
jrriehl, line 58: Please use Google style docstrings as we do in the rest of the main library:
```
"""
Description

Args:
   arg (type): description

Returns:
    type: description
"""
```

libs/core/uagents_core/utils/registration.py
```diff
@@ -0,0 +1,229 @@
...
+    Register the agent with the Agentverse API.
+    :param identity: The identity of the agent.
+    :param url: The URL endpoint for the agent
+    :param protocol_digest: The digest of the protocol that the agent supports
+    :param agentverse_token: The token to use to authenticate with the Agentverse API
+    :param agent_title: The title of the agent
+    :param readme: The readme for the agent
+    :param agentverse_url: The URL of agentverse (if different from the default)
+    :return:
+    """
```
jrriehl, line 107: Please use Google style docstrings.

libs/core/uagents_core/utils/communication.py
```diff
@@ -0,0 +1,87 @@
...
+from uagents_core.envelope import Envelope
+from uagents_core.config import DEFAULT_AGENTVERSE_URL, DEFAULT_ALMANAC_API_PATH
+from uagents_core.logger import get_logger
+
+
+
+logger = get_logger("uagents_core.utils.communication")
+
+
+def lookup_endpoint_for_agent(agent_address: str, *, agentverse_url: Optional[str] = None) -> str:
```
jrriehl, line 18: This is a simplified version of the resolver, which does not handle multiple endpoints, for example. Not sure whether it makes sense to include here.

jrriehl (commented):
Looks good overall. My main comments are about the utilities. Also this should probably go under `python/libs`, right?

libs/core/uagents_core/logger.py
qati: I wanted to introduce unified logging and log style to be adopted in the future.

libs/core/uagents_core/utils/communication.py
qati: first version is fine as it is I guess, but follow up versions needs to improve and add support for multiple endpoints.

libs/core/pyproject.toml
```diff
@@ -0,0 +1,67 @@
...
+bech32 = "^1.2.0"
+ecdsa = "^0.19.0"
+aiohttp = "^3.8.3"
+requests =">=2.32.3,<3.0"
+structlog = "^24.4.0"
+rich = "^13.9.4"
+
+
+[tool.poetry.group.dev.dependencies]
+black = "^24.10.0"
```
Archento, line 29: black is not used in this project and the lib right now does not have any tests so please remove all unneeded packages if this is meant to have its own pyproject file.

.github/CODEOWNERS
```diff
@@ -1,5 +1,5 @@
 # The default code owners of the uagents repo.
-* @jrriehl @Archento @lrahmani
+* @jrriehl @Archento @lrahmani @qati
```
Archento, line 2:
```suggestion
* @jrriehl @Archento @lrahmani

/libs/core/ @qati
```

I propose this scope as your PR only targets changes in this directory.

libs/core/README.md
Archento, line 1: Please add a description here even if it is preliminary to understand the reasoning behind this package / its reason for existence.

libs/core/pyproject.toml
```diff
@@ -0,0 +1,67 @@
+[tool.poetry]
+name = "uagents-core"
+version = "0.1.0"
+description = "Lightweight framework for rapid agent-based development"
```
Archento, line 4: Please check all fields in this file as this seems to be a copy-paste error.

Archento (changes requested):
Since there is a lot of copied/moved code it is actually quite hard to identify what you've changed or added. Also there is no description added to the PR so I can't really understand the reasoning for adding this library.

What is the main benefit for this structure compared to importing what you need for a project directly from uagents?

libs/core/uagents_core/utils/registration.py
```diff
@@ -0,0 +1,278 @@
...
+        "agent_address": agent_address,
+        "agent_endpoint": request.endpoint or "",
+        "protocol_digest": ",".join(protocol_digests),
+    }
+    if request.endpoint is None:
+        logger.warning(
+            "No endpoint provided for agent registration",
+            extra=registration_metadata,
+        )
+        return
```
jrriehl, line 126: For mailbox and proxy agents, we should probably use the standard fixed endpoints: https://github.com/fetchai/uAgents/blob/main/python/src/uagents/config.py#L106-L109.

libs/core/pyproject.toml
jrriehl: Some additional context is that we use `ruff` with pre-commit hooks in the main project and it works very well for us. Maybe worth considering adopting here as well.

PR #604, opened 2025-01-09
fix(core): enable configuring RegistrationPolicy on agent creation

Dacksus:
Before this fix, RegistrationPolicies couldn't be specified when initializing an agent because they would require the agent's identity as an argument, which is available only after initialization. This fix removes the need to manually specify an identity and lets the framework take care of it.


jrriehl (dismissed):
LGTM

Archento (approved):
All good from my side 👍🏻

jrriehl (approved):
LGTM

Issue #601, opened 2025-01-07
Enhance robustness to ledger RPC errors


jrriehl:
Under normal operations, ledger transactions and queries are reasonably fast and do not disrupt other agent functionality, but when errors occur, agents can become unresponsive. Since the interactions related to Almanac registration are not time critical, we should make sure that these do not block other agent functions when the ledger is not responding.

Dacksus:
I didn't find any occasion where grpc/legder issues would render the agent unresponsive, but wasn't able to trigger all possible exceptions, yet.

PR #610, opened 2025-01-17
chore(core): improve logging control and experience

Dacksus:
## Proposed Changes

Specifying a log_level on agent initialization will apply this log level to all loggers.
Started moving exception details to the DEBUG level for a cleaner "normal" output.

## Linked Issues

#601 

## Types of changes

Refactorings. (No application logic touched)

### DIFF
--- a/python/src/uagents/agent.py
+++ b/python/src/uagents/agent.py
@@ -83,7 +83,7 @@ from uagents.types import (
     RestMethod,
     RestPostHandler,
 )
-from uagents.utils import get_logger
+from uagents.utils import get_logger, set_global_log_level
 
 
 async def _run_interval(
@@ -329,6 +329,8 @@ class Agent(Sink):
 
         # initialize wallet and identity
         self._initialize_wallet_and_identity(seed, name, wallet_key_derivation_index)
+        if log_level != logging.INFO:
+            set_global_log_level(log_level)
         self._logger = get_logger(self.name, level=log_level)
 
         self._agentverse = parse_agentverse_config(agentverse)
@@ -379,7 +381,6 @@ class Agent(Sink):
                 self._wallet,
                 self._almanac_contract,
                 self._test,
-                logger=self._logger,
                 almanac_api=self._almanac_api_url,
             )
         self._metadata = self._initialize_metadata(metadata)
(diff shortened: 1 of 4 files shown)


python/src/uagents/agent.py
```diff
@@ -726,6 +727,7 @@ def sign_digest(self, digest: bytes) -> str:
         """
         return self._identity.sign_digest(digest)
 
+    # TODO this is not used anywhere in the framework
```
Dacksus, line 730: blatant piggy-back on this PR. Are we aware that this is currently unused? Can we remove it or do we want to keep it for "backward compatibility"?

python/src/uagents/network.py
```diff
@@ -528,15 +530,16 @@ def query_contract(self, query_msg: Dict[str, Any]) -> Any:
             Any: The query response.
 
         Raises:
-            RuntimeError: If the contract address is not set or the query fails.
+            ValueError: If the response from contract is not a dict.
```
Dacksus, line 533: Seemed to be an outdated docstring, as no RuntimeError is raised anymore

python/src/uagents/utils.py
```diff
@@ -35,4 +35,19 @@ def log(logger: Optional[logging.Logger], level: int, message: str):
...
+    Set the log level for all modules globally. Can still be overruled manually.
+
+    Args:
+        level (Union[int, str]): The logging level as defined in _logging_.
+    """
+    logging.basicConfig(level=level)
+    # the manager of the root logger should be its only instance and contain
+    # all loggers of the process. Update the new log level in all of them.
+    for name in logging.Logger.manager.loggerDict:
+        # TODO only apply to framework loggers?
```
Dacksus, line 49: This currently sets the log level for all loggers including those of other libraries as all loggers are collected under one root logger. Any preferences for adding a filter to only apply it to uagents modules?
Personally, I think it's okay this way, as it will only get into effect in those rare cases when log level is increased for debugging and under normal conditions nothing happens.

jrriehl (approved):
LGTM

python/src/uagents/agent.py
jrriehl: This is used in the example here: https://github.com/fetchai/uAgents/blob/main/python/examples/07-msg-verification/main.py#L29, which is not a particularly useful example, but I guess we should keep it at least until we drop this example.

PR #612, opened 2025-01-21
feat(core): allow general access to rest endpoints, except inspector ones

Dacksus:
This PR makes all agent REST endpoints available to any caller, except those used for the inspector which will still be restricted to localhost only.


Archento (dismissed):
LGTM!

jrriehl (changes requested):
Looks good, just needs a test fix.

jrriehl (approved):

PR #639, opened 2025-02-24
feat(core): add send and receive function

jrriehl:
## Proposed Changes

Add new function to send and wait for response to a message. In this case, the sender must specify the response type, and can also specify a timeout. Works for both async and sync messaging.

## Types of changes

- [ ] Documentation update.

## Checklist

### If applicable

### DIFF
--- a/python/src/uagents/context.py
+++ b/python/src/uagents/context.py
@@ -22,6 +22,7 @@ from typing import (
 
 import requests
 from cosmpy.aerial.client import LedgerClient
+from pydantic import ValidationError
 from typing_extensions import deprecated
 
 from uagents.communication import (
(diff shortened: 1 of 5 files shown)


Dacksus (dismissed):
Nice and clean! Very elegant realization for this long awaited feature!

python/src/uagents/context.py
```diff
@@ -525,9 +554,6 @@ async def send_raw(
                 session=self._session,
             )
 
-        if isinstance(result, Envelope):
```
jrriehl, line 528: This seemed to be redundant, as we already call it in the dispenser.

python/src/uagents/context.py
Archento, line 1: the `.send()` in broadcast still has the sync argument

Archento (changes requested):
Awesome to finally land this! Great work :)

I've attached a couple minor remarks

Archento (approved):

Archento (approved):

PR #646, opened 2025-03-03
feat(core): add protocol specification

jrriehl:
## Proposed Changes

Add protocol specification for easier definition of protocols based on interactions and roles. The original way of defining protocols still works exactly the same, but now if you provide a protocol specification object, for example:

```python
from uagents import Model


class ProposeChat(Model):
    pass


class AcceptChat(Model):
    pass


class RejectChat(Model):
    pass


class Chat(Model):
    text: str


class EndChat(Model):
    pass


spec = ProtocolSpecification(
    interactions={
        ProposeChat: {AcceptChat, RejectChat},
        AcceptChat: {Chat, EndChat},
        RejectChat: set(),
        Chat: {Chat, EndChat},
        EndChat: set(),
    },
    roles={
        "user": {ProposeChat, Chat, EndChat},
        "agent": {AcceptChat, RejectChat, Chat, EndChat},
    },
)
```

Then you can generate the protocols for both roles: 

```python
user_proto = Protocol(name="ChatProtocol", version="0.1", spec=spec, role="user")
agent_proto = Protocol(name="ChatProtocol", version="0.1", spec=spec, role="agent")
```

Creating the protocol this way locks the protocol so that you cannot add handlers for messages that are not already included int he interactions, ensuring that the protocol manifest and digest do not change. Also, when an agent includes a protocol, it calls the `protocol.verify()` function, which checks that all the required handlers for the role have been implemented.

## Linked Issues

## Types of changes

- [ ] Documentation update.

## Checklist

### If applicable


python/examples/21-protocol-interactions/main.py
Archento, line 1: This not really needed I think if we cover the possible outcomes in the test so I'd rather see this file with a bit more comments/explanation in the Examples repo.

python/src/uagents/protocol.py
```diff
@@ -19,7 +56,13 @@ class Protocol:
 
     """
 
-    def __init__(self, name: Optional[str] = None, version: Optional[str] = None):
+    def __init__(
+        self,
+        name: Optional[str] = None,
+        version: Optional[str] = None,
+        spec: Optional[ProtocolSpecification] = None,
```
Archento, line 63: `None` for backwards compatibility?

Archento (commented):
I really like this comprehensive approach of defining a protocol. Ultimately we'd benefit a  lot from separating the syntax of the protocol from its implementation.
Great work! 👍🏻

python/src/uagents/protocol.py
jrriehl: Yes, it's not meant to be the required way to create a protocol, just an option.

python/examples/21-protocol-interactions/main.py
jrriehl: Agreed, removed.

python/src/uagents/protocol.py
```diff
@@ -37,6 +80,12 @@ def __init__(self, name: Optional[str] = None, version: Optional[str] = None):
         self._version = version or "0.1.0"
         self._canonical_name = f"{self._name}:{self._version}"
         self._digest = ""
+        self._interactions: Optional[Dict[Type[Model], Set[Type[Model]]]] = None
```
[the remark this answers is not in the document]
jrriehl: Good point, you are right that there is some overlap here. The main difference is the model digests. I'll see if I can nicely remove `_replies` by adding the digests to interactions.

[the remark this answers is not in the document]
jrriehl: It turned out to be easier to get rid of self._interactions instead. :smile:

Archento (dismissed):
Looks good to me now!

Archento (approved):

PR #647, opened 2025-03-04
Feat/search agent example

Dacksus:
This PR provides an example for how to implement a search agent to extend on the basic agentverse search functionality. The search agent will get a general list for agents potentially satisfying the search request from agentverse search engine and then verify additional agent specific attributes with each agent individually. This allows for example to search for all EV charging stations in a certain area (search engine) that are below a given price limit (search agent).

There are two implementations for the search agent: The async implementation uses the common message handler-based approach and potentially scales better. The sync implementation makes use of the new `send_and_receive` feature, which makes it easier to follow the code for people unfamiliar with asynchronous communication.


python/examples/21-search/charger_search_agent_sync.py
```diff
@@ -0,0 +1,87 @@
...
+    query = AttributeQuery(
+        search_id=search_session, attributes=list(msg.attribute_filter)
+    )
+
+    responses = {}
+    for agent in agents:
+        ctx.logger.info(f"querying agent: {agent.address}")
+        reply: AttributeResponse
+        reply, status = await ctx.send_and_receive(
+            agent.address, query, response_type=AttributeResponse
```
jrriehl, line 56: Maybe worth specifying the timeout here at least to make it more easily configurable? I think the default is 60s.

Dacksus: Agree to explicitly define the timeout to make it more visible for the example, but would you include it as a field in the protocol?

jrriehl: I probably wouldn't. I was just thinking about setting it to something lower to not get hung up for too long on unresponsive agents.

beshr11 (approved):

PR #659, opened 2025-03-27
feat(core): add agent chat protocol

jrriehl:
## Proposed Changes

- Adds agent chat protocol to new `uagents_core.contrib` module
- Fixes import bug in `uagents_core.utils`
- Updates reply validation to log an error if at least one reply required by the protocol was not sent

## Linked Issues

## Types of changes

- [ ] Documentation update.

## Checklist

### If applicable

## Further comments

### DIFF
--- a/python/src/uagents/agent.py
+++ b/python/src/uagents/agent.py
@@ -1328,6 +1328,8 @@ class Agent(Sink):
                 except Exception as ex:
                     self._logger.exception(f"Exception in message handler: {ex}")
 
+            context.validate_replies()
+
 
 class Bureau:
     # pylint: disable=protected-access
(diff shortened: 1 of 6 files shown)


Archento (commented):
I see that this is now checked for every sent message and the function will only log if an outbound message was not a valid reply.

I think this logic currently will also log if an error message is sent from a context as this would use the same session id. Also the chained case is not handled as an  outgoing message can also be sent to a different agent.

jrriehl:
> I see that this is now checked for every sent message and the function will only log if an outbound message was not a valid reply.
Yes, I think we could definitely augment this with validation of outgoing / starter messages. 

> I think this logic currently will also log if an error message is sent from a context as this would use the same session id. Also the chained case is not handled as an outgoing message can also be sent to a different agent.
Good point, we should allow for ErrorMessage by default here, too. I'll make an update.

Archento (approved):
LGTM!

PR #661, opened 2025-03-28
feat(core): add agent subscription protocol spec

Archento:
## Proposed Changes

Add the agent subscription protocol spec to uagents_core to allow agents to accept subscribers.
Also fix minor typos.

## Types of changes

- [x] Documentation update.

## Checklist

### If applicable

### DIFF
--- a/python/uagents-core/uagents_core/protocol.py
+++ b/python/uagents-core/uagents_core/protocol.py
@@ -9,12 +9,23 @@ from uagents_core.models import Model
 
 
 class ProtocolSpecification(BaseModel):
-    """Specification for the interactions and roles of a protocol."""
+    """
+    Specification for the interactions and roles of a protocol.
 
-    interactions: dict[type[Model], set[type[Model]]]
-    roles: dict[str, set[type[Model]]] | None = None
+    Args:
+        name (str): The name of the protocol.
+        version (str): The version of the protocol.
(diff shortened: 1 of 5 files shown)


python/uagents-core/uagents_core/protocol.py
```diff
@@ -9,12 +9,23 @@
...
+    Specification for the interactions and roles of a protocol.
+
+    Args:
+        name (str): The name of the protocol.
+        version (str): The version of the protocol.
+        interactions (dict[type[Model], set[type[Model]]]): A mapping of models
+            to the set of models that can reply to them.
+        roles (dict[str, set[type[Model]]] | None): A mapping of role names to
+            the set of models that need to be implemented by that role.
+            If None, all models can be used in all roles.
```
jrriehl, line 22: Minor suggestion for clarity:
```suggestion
    Args:
        name (str): The name of the protocol.
        version (str): The version of the protocol.
        interactions (dict[type[Model], set[type[Model]]]): A mapping of models
            to the corresponding set of models that are valid replies.
        roles (dict[str, set[type[Model]]] | None): A mapping of role names to the set of
            incoming message models that role needs to be implement handlers for.
            If None, all models can be implemented by all roles.
```

jrriehl (commented):
Nice one, just a suggestion on the doc string.

PR #679, opened 2025-04-07
feat(integration): Adding langchain and crewai adapter for uAgents

abhifetch:
## Proposed Changes

Addititon of langchain uagents adapter in experiment

## Linked Issues

## Types of changes

- [ ] Documentation update.

## Checklist

### If applicable

## Further comments


abhifetch:
can find example and readme for how to use it here : https://github.com/abhifetch/uAgents-Adapter/settings

Archento (changes requested):
Thank you for your contribution.

I strongly believe that this belongs in the `3-applications` folder of the official [uAgents examples repo](https://github.com/fetchai/uAgent-Examples) as it introduces dependencies into the core framework and is more of an integration than a core functionality of uagents.
Also, the target folder here would be `/contrib` rather than `/experimental`

abhifetch:
@Archento Thanks for the suggestion, I have discussed this with @ejfitzgerald and got this into the experimental folder. I know this is integration, we want developers to use these integrations and directly get their agent on agentverse with chat protocol enabled.

@ejfitzgerald can you please comment on what should be preferred here.

python/src/uagents/experimental/adapters/crewai/tools.py
```diff
@@ -0,0 +1,780 @@
...
+import threading
+import time
+from datetime import datetime
+from threading import Lock
+from typing import Any, Dict, List
+from uuid import uuid4
+
+import requests
+from langchain_core.callbacks import CallbackManagerForToolRun
+from langchain_core.tools import BaseTool
```
qati, line 15: if uagents is installed, and we import in this from experimental, and the dev doesn't have langchain_core this will just crash

python/src/uagents/experimental/adapters/crewai/tools.py
```diff
@@ -0,0 +1,780 @@
...
+
+# Helper function to extract parameters from text using OpenAI's GPT-4o model with the Chat Completions API
+def extract_params_from_text(text: str, param_keys: List[str]) -> Dict[str, str]:
+    """
+    Extract parameters from text using OpenAI's GPT-4o model with the Chat Completions API
+    """
+    import json
+    import os
+
+    from openai import OpenAI
```
qati, line 58: how we know correct version of openai package exist?

jrriehl:
These look really cool, but as others have already pointed out, `openai` and `langchain_core` are not dependencies of `uagents`, so these will fail without an extra installation step. I feel like these could go directly in https://github.com/fetchai/uAgent-Examples and would be a perfect git there as integrations.

python/src/uagents/experimental/adapters/crewai/tools.py
abhifetch: updated and created a package in itself for adapters

python/src/uagents/experimental/adapters/crewai/tools.py
abhifetch: thankyou, yes it was a perfect choice to create package instead

python/uagents-adapter/pyproject.toml
```diff
@@ -0,0 +1,66 @@
...
+optional = true
+
+[tool.poetry.group.langchain.dependencies]
+langchain-core = ">=0.3.52,<0.4.0"
+langchain-openai = ">=0.3.13,<0.4.0"
+
+[tool.poetry.group.crewai]
+optional = true
+
+[tool.poetry.group.crewai.dependencies]
```
qati, line 34: given if its installed as a package both can be imported in, I would just go with simple flat dependencies for easiness, and then ci doesn't need much update to do auto pypi release. what you think @jrriehl ?

python/src/uagents/config.py
```diff
@@ -29,7 +29,7 @@
 REGISTRATION_RETRY_INTERVAL_SECONDS = 60
 AVERAGE_BLOCK_INTERVAL = 6
 DEFAULT_LEDGER_TX_WAIT_SECONDS = 30
-ALMANAC_CONTRACT_VERSION = "2.0.0"
```
qati, line 32: do we need to do this change here?

python/src/uagents/context.py
```diff
@@ -496,10 +496,14 @@ async def send_raw(
                     session=self._session,
                 )
 
-            self._outbound_messages[parsed_address] = (
-                message_body,
-                message_schema_digest,
-            )
+            if parsed_address in self._outbound_messages:
```
qati, line 499: not sure about these bits?

python/src/uagents/network.py
```diff
@@ -210,7 +210,12 @@ def check_version(self) -> bool:
         """
         try:
             deployed_version = self.get_contract_version()
-            if deployed_version != ALMANAC_CONTRACT_VERSION:
+
+            # parse major and minor versions
+            deployed_major_version = deployed_version.split(".")[0]
+            supported_major_version = ALMANAC_CONTRACT_VERSION.split(".")[0]
```
qati, line 216: do we need to change this?

python/src/uagents/config.py
abhifetch: yes, it is as per the latest uagents version. The PR was originally created earlier.It should be 2.1.0

[here](https://github.com/fetchai/uAgents/blob/main/python/src/uagents/config.py)

python/src/uagents/context.py
abhifetch: again i have not made these changes explicitly, this is as per latest uagents version.

python/src/uagents/network.py
abhifetch: yes

python/uagents-adapter/examples/trip_planner/pyproject.toml
```diff
@@ -0,0 +1,32 @@
+[tool.poetry]
+name = "trip-planner-crew"
+version = "0.1.0"
+description = ""
+authors = ["Your Name <you@example.com>"]
+
+[tool.poetry.dependencies]
+python = ">=3.10.0,<3.12"
+crewai = "^0.11.0"
```
qati, line 9: this example will be updated after the pypi release of uagents-adaptors?

qati (approved):

python/uagents-adapter/src/uagents_adapter/crewai/tools.py
```diff
@@ -0,0 +1,760 @@
...
+        thread = threading.Thread(target=run_agent, daemon=True)
+        thread.start()
+        agent_info["thread"] = thread
+
+        # Give the agent a moment to start
+        time.sleep(0.5)
+
+        return agent_info
+
+    def _register_agent_with_agentverse(self, agent_info):
```
jrriehl, line 488: It might be a good idea to import the `uagents_core` utility that does this: https://github.com/fetchai/uAgents/blob/main/python/uagents-core/uagents_core/utils/registration.py#L138.

jrriehl: Nevermind, I see this is hitting the agent endpoint, which makes sense.

jrriehl (approved):

python/uagents-adapter/src/uagents_adapter/crewai/tools.py
abhifetch: cool thanks.

Archento (approved):
A couple of things here:
- why is the register tool not a separate class that both derivates can use
- why do you use threading when most of the current architecture can make better use of multiprocessing or even simply asyncio tasks in a single thread
- adding to the above: hardcoded time.sleep should generally be avoided and instead sequential event based approaches used  
- the chat handlers have overly complex case logic which is prone to errors and could potentially be simplified
- the pyproject file could be updated to adhere to the current python standard which is tool agnostic (avoid, if possible, that Poetry needs to be used)
- pytest is imported but there are no tests
- since >=python3.10 is defined, please also adhere to the new typing requirements

All of these should eventually be tackled but I assume this can happen down the line / as a patch afterwards.

Otherwise looks good to me 👍🏻

abhifetch:
Thankyou @Archento , I will address all of these in the coming release.

PR #684, opened 2025-04-08
fix(core): reply validation

jrriehl:
## Proposed Changes

Fixes erroneous logging errors on reply validation:
- Changes outbound message to a list in case multiple messages are sent to the same agent
- When replies are defined as empty, do not validate replies

## Types of changes

- [ ] Documentation update.

## Checklist


Archento (dismissed):
LGTM

PR #687, opened 2025-04-10
chore(core): update almanac contract version and check

jrriehl:
## Proposed Changes

Update almanac contract version to latest deployed version and relax version check.

## Types of changes

- [ ] Documentation update.

## Checklist

### If applicable


MissingNO57 (approved):
LGTM

python/src/uagents/network.py
```diff
@@ -210,7 +210,12 @@ def check_version(self) -> bool:
         """
         try:
             deployed_version = self.get_contract_version()
-            if deployed_version != ALMANAC_CONTRACT_VERSION:
+
+            # parse major and minor versions
+            deployed_major_version = deployed_version.split(".")[0]
+            supported_major_version = ALMANAC_CONTRACT_VERSION.split(".")[0]
+
+            if supported_major_version != deployed_major_version:
```
pbukva, line 218: I'm not quite sure if returning the `False` in this case (the [line 224](https://github.com/fetchai/uAgents/pull/687/files#diff-7ec41fe6bd4bde576f5d7db4e27c06b82e7673088c6dfa9ff298b4f3540916afR224)) is the right way to go.

pbukva: And if yes, I would suggest to put the value of the `supported_major_version` in to the newtork config, so users can change it if they would need to ...

PR #699, opened 2025-05-01
fix(core): set default replies to empty set

jrriehl:
## Proposed Changes

Previously there was a difference in meaning between specifying `replies=None` and `replies=set()` because we enforced that no replies should be sent in the latter but not the former case. This is no longer the case as of https://github.com/fetchai/uAgents/pull/684, so it makes sense to simply set the default to an empty set.

Setting `replies=None` also had the effect of excluding the message from the set of models that define the manifest. This meant that when building a protocol with handlers without specifying replies, the manifest and resulting digest were incorrect.

## Types of changes

- [ ] Documentation update.

## Checklist

### If applicable


Archento (approved):
LGTM!

PR #703, opened 2025-05-06
chore(core): return registration tx response

jrriehl:
## Proposed Changes

Return registration transaction status and log hash on error. Also exposes tirneout height and gas limit parameters.

## Types of changes

- [ ] Documentation update.

## Checklist

### DIFF
--- a/python/src/uagents/network.py
+++ b/python/src/uagents/network.py
@@ -445,7 +445,7 @@ class AlmanacContract(LedgerContract):
         broadcast_retry_delay: RetryDelayFunc | None = None,
         poll_retries: int | None = None,
         poll_retry_delay: RetryDelayFunc | None = None,
-    ) -> None:
+    ) -> TxResponse:
         """
         Register an agent with the Almanac contract.
 
@@ -456,6 +456,9 @@ class AlmanacContract(LedgerContract):
             protocols (list[str]): List of protocols.
             endpoints (list[dict[str, Any]]): List of endpoint dictionaries.
             signature (str): The agent's signature.
+
+        Returns:
+            TxResponse: The transaction response.
         """
         if not self.address:
             raise ValueError("Contract address not set")
@@ -509,7 +512,11 @@ class AlmanacContract(LedgerContract):
             poll_retry_delay=poll_retry_delay,
         )
         if status.code != 0:
-            raise RuntimeError("Registration transaction failed")
+            raise RuntimeError(
+                f"Registration transaction failed ({status.code}): {status.hash})"
+            )
+
+        return status
 
     async def register_batch(
         self,
@@ -521,7 +528,7 @@ class AlmanacContract(LedgerContract):
         broadcast_retry_delay: RetryDelayFunc | None = None,
         poll_retries: int | None = None,
         poll_retry_delay: RetryDelayFunc | None = None,
-    ) -> None:
+    ) -> TxResponse:
         """
         Register multiple agents with the Almanac contract.
 
@@ -529,6 +536,9 @@ class AlmanacContract(LedgerContract):
             ledger (LedgerClient): The Ledger client.
             wallet (LocalWallet): The wallet of the registration sender.
             agents (list[ALmanacContractRecord]): The list of signed agent records to register.
+
+        Returns:
+            TxResponse: The transaction response.
         """
         if not self.address:
             raise ValueError("Contract address not set")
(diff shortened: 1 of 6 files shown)


lrahmani (dismissed):
LGTM

python/src/uagents/network.py
```diff
@@ -49,6 +49,7 @@
 RetryDelayFunc = Callable[[int], float]
 DEFAULT_BROADCAST_RETRIES = 5
 DEFAULT_POLL_RETRIES = 10
+DEFAULT_REGISTRATION_TIMEOUT_BLOCKS = 100
```
Archento, line 52: Depending on the speed of the network a block is created roughly every 5-6 seconds. This means that the timeout for a valid registration request would be around 8-10 minutes if we want this constant to be `100`.

python/src/uagents/network.py
```diff
@@ -488,12 +501,18 @@ async def register(
...
         tx: SubmittedTx | None = None
         for n in range(num_broadcast_retries):
             try:
                 tx = prepare_and_broadcast_basic_transaction(
-                    ledger, transaction, wallet, account=account
+                    ledger,
+                    transaction,
+                    wallet,
+                    account=account,
+                    gas_limit=gas_limit,
```
Archento, line 514: Gas limit makes sense here because we don't want the registration to go through if it is too expensive. This does mean though that the gas limit should be configurable somewhere which I haven't seen yet.

python/src/uagents/network.py
```diff
@@ -488,12 +501,18 @@ async def register(
...
+                    timeout_height=timeout_height,
```
Archento, line 515: Using the timeout_height here does not really make sense if the structure of the registration is the same as before. To use it effectively we will need to refactor this method to:
-  fire a single transaction with a certain block height,
- check if the registration was successful, and
- retry with a new transaction if not.

This way we would be using the least amount of tokens for each registration attempt which is ultimately what we should be aiming for.

python/src/uagents/network.py
```diff
@@ -509,7 +528,11 @@ async def register(
             poll_retry_delay=poll_retry_delay,
         )
         if status.code != 0:
-            raise RuntimeError("Registration transaction failed")
+            raise RuntimeError(
+                f"Registration transaction failed ({status.code}): {status.hash})"
+            )
+
+        return status
```
Archento, line 535: What's the benefit of returning the whole transaction here?

python/src/uagents/network.py
```diff
@@ -899,8 +947,13 @@ async def register(
...
+        timeout_height = ledger.query_height() + timeout_blocks
         submitted_transaction: SubmittedTx = prepare_and_broadcast_basic_transaction(
-            client=ledger, tx=transaction, sender=wallet
+            client=ledger,
+            tx=transaction,
+            sender=wallet,
+            gas_limit=gas_limit,
+            timeout_height=timeout_height,
         )
         await wait_for_tx_to_complete(submitted_transaction.tx_hash, ledger)
```
Archento, line 958: This can fail but it isn't caught (especially together with the timeout height)

python/src/uagents/network.py
jrriehl: Yes, that was my intention to leave this relatively long as a default.

python/src/uagents/network.py
jrriehl: You mean instead of just the hash? I did it this way because `TxResponse` seems to be a logical return type for a method whose main purpose is to execute a transaction and it contains the hash and other metadata that could potentially be useful.

python/src/uagents/network.py
jrriehl: Good point, we can make this configurable in the ledger registration policies.

python/src/uagents/network.py
jrriehl: Okay, I added a new timeout height with each retry.

python/src/uagents/network.py
Archento: Alright, makes sense to me :)

python/src/uagents/network.py
Archento: I think my issue with the current implementation is not yet resolved. Adding the `timeout_height` here does not provide meaningful benefit yet(!). It just ensures that transactions that are not handled by a node will be dropped if picked up after the height is reached but if we still broadcast multiple transactions a number of times and not check them individually we'll end up spending more tokens than necessary.

What I'm trying to say is that the current implementation is not breaking anything but at the same time it would require another logic altogether to use the least amount of tokens for a registration.

python/src/uagents/network.py
Archento: This ties into my other comment: if we check regularly for failed transactions I'd expect this timeout window to be quite narrow 😄

python/src/uagents/network.py
jrriehl: Issue added here: https://github.com/fetchai/uAgents/issues/705.

PR #728, opened 2025-06-11
feat(core): Add register_batch_in_almanac function

AaronJamesKing:
## Proposed Changes

Adds a `register_batch_in_almanac` function to registration utils, which utilizes the Almanac's [batch registration endpoint](https://github.com/fetchai/api-clients/blob/main/agentverse-client/agentverse_client/almanac/docs/AgentRegistrationAttestationBatch.md)

## Linked Issues

N/A

## Types of changes

- [ ] Documentation update.

## Checklist

### If applicable

## Further comments

Because the `register_batch_in_almanac` function does not have unit tests or public-facing documentation, I didn't add any for this new function either. ~What is the recommended way to manually perform end-to-end tests for a new function like this?~ Changes were tested manually in a Python REPL.

### DIFF
--- a/python/uagents-core/uagents_core/utils/registration.py
+++ b/python/uagents-core/uagents_core/utils/registration.py
@@ -28,7 +28,7 @@ from uagents_core.registration import (
     RegistrationRequest,
     RegistrationResponse,
 )
-from uagents_core.types import AgentEndpoint
+from uagents_core.types import AddressPrefix, AgentEndpoint
 
 logger = get_logger("uagents_core.utils.registration")
 
@@ -45,7 +45,7 @@ class AgentRegistrationInput:
         identity: Identity,
         endpoints: list[str],
         protocol_digests: list[str],
-        prefix: str | None = None,
+        prefix: AddressPrefix | None = None,
         metadata: dict[str, str | list[str] | dict[str, str]] | None = None,
     ):
         self.identity = identity
@@ -111,7 +111,7 @@ def register_in_almanac(
     endpoints: list[str],
     protocol_digests: list[str],
     metadata: dict[str, str | list[str] | dict[str, str]] | None = None,
-    prefix: str | None = None,
+    prefix: AddressPrefix | None = None,
     *,
     agentverse_config: AgentverseConfig | None = None,
     timeout: int = DEFAULT_REQUEST_TIMEOUT,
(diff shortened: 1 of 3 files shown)


jrriehl (approved):
LGTM. There could be a case to make for registering the partial batch if only one or a few agents fail, but it might be cleanest to keep it all or nothing.

Archento (approved):
I agree with @jrriehl that the batch function could benefit from adding a "partial" flag and should then return a mapping of which agent may have failed.
Otherwise LGTM 👍🏻

AaronJamesKing:
👍 Added a `validate_all_before_registration` flag, and changed the return value to a tuple that includes whether registration succeeded as well as the list of identities that failed validation.

Archento (approved):
Works for me 👍🏻

Archento:
Please only open Pull Requests if your changes are final. Otherwise you can state somewhere that this is a work-in-progress or ideally open a **Draft** Pull Request instead.
Thanks

jrriehl (approved):

python/uagents-core/uagents_core/utils/registration.py
```diff
@@ -77,6 +121,8 @@ def register_in_almanac(
 
     Args:
         identity (Identity): The identity of the agent.
+        prefix (str | None): The Almanac contract where the agent will be registered.
+            If none is provided, the agent will be registered to the default contract.
```
jrriehl: Suggest simplifying this since registering to Alamanc API does not actually require any contract interactions.
```suggestion
        prefix (AddressPrefix | None): The prefix for the agent identifier.
```

python/uagents-core/uagents_core/utils/registration.py
```diff
@@ -63,11 +86,32 @@ def _send_post_request(
...
+    attestation.sign(item.identity)
+    return attestation
+
+
 def register_in_almanac(
     identity: Identity,
     endpoints: list[str],
     protocol_digests: list[str],
     metadata: dict[str, str | list[str] | dict[str, str]] | None = None,
+    prefix: str | None = None,
```
jrriehl:
```suggestion
    prefix: AddressPrefix | None = None,
```

python/uagents-core/uagents_core/utils/registration.py
```diff
@@ -32,6 +33,28 @@
...
+    endpoints: list[str]
+    protocol_digests: list[str]
+    metadata: dict[str, str | list[str] | dict[str, str]] | None = None
+
+    def __init__(
+        self,
+        identity: Identity,
+        endpoints: list[str],
+        protocol_digests: list[str],
+        prefix: str | None = None,
```
jrriehl:
```suggestion
        prefix: AddressPrefix | None = None,
```

jrriehl (changes requested):
I realized that we should actually be a bit more strict on what is allowed for the prefix.

And event better would probably be to import this from `uagents_core.types`: https://github.com/fetchai/uAgents/blob/main/python/uagents-core/uagents_core/types.py#L10.

PR #737, opened 2025-06-25
feat(experimental): add basic subscribable protocol

jrriehl:
## Proposed Changes

Adds a basic version of a subscribable protocol that checks the subscription tier of an agent before fulfilling the request. It is tightly based on the existing `QuotaProtocol` which establishes a free tier that is bypassed for agents on a paid tier.

### Usage examples

```python
from uagents.experimental.subscription import SubscribableProtocol

# Initialize the SubscribableProtocol instance
subs_protocol = SubscribableProtocol(
    storage_reference=agent.storage,
    identity=agent.identity,
    agentverse=agent.agentverse,
    name="quota_proto",
    version=agent._version,
    # default_rate_limit=RateLimit(window_size_minutes=1, max_requests=3), # Optional
)

# Subscription and rate limiting does not apply to this message handler
@subs_protocol.on_message(ExampleMessage1)
async def handle(ctx: Context, sender: str, msg: ExampleMessage1):
    ...

# This message handler is rate limited with custom window size and request limit
# that will be bypassed if the agent is on a paid tier
@subs_protocol.on_message(
    ExampleMessage2,
    rate_limit=RateLimit(window_size_minutes=1, max_requests=3),
)

async def handle(ctx: Context, sender: str, msg: ExampleMessage2):
    ...
```

## Types of changes

- [ ] Documentation update.

## Checklist

### If applicable

### DIFF
--- a/python/src/uagents/experimental/subscription/__init__.py
+++ b/python/src/uagents/experimental/subscription/__init__.py
@@ -1,0 +1,108 @@
+"""
+This Protocol class provides a basic example to support subscriptions.
+
+If a requesting agent is on the free tier, it will be rate-limited according
+to the specification given for the QuotaProtocol. If the agent is on a paid
+tier (PLUS or PRO), it will not be rate-limited and can make requests
+without restrictions.
+
+Usage examples:
+
+```python
+from uagents.experimental.subscription import SubscribableProtocol
+
+# Initialize the SubscribableProtocol instance
+subs_protocol = SubscribableProtocol(
+    storage_reference=agent.storage,
+    identity=agent.identity,
+    agentverse=agent.agentverse,
+    name="quota_proto",
+    version=agent._version,
+    # default_rate_limit=RateLimit(window_size_minutes=1, max_requests=3), # Optional
+)
+
+# Subscription and rate limiting does not apply to this message handler
+@subs_protocol.on_message(ExampleMessage1)
+async def handle(ctx: Context, sender: str, msg: ExampleMessage1):
+    ...
+
+# This message handler is rate limited with custom window size and request limit
+# that will be bypassed if the agent is on a paid tier
+@subs_protocol.on_message(
+    ExampleMessage2,
+    rate_limit=RateLimit(window_size_minutes=1, max_requests=3),
+)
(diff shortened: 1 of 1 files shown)


Dacksus:
I wonder if it would make sense to merge it with the general [Subscription protocol](https://github.com/fetchai/uAgents/blob/be012455e3e622c8c4a055e415bd2732898ce362/python/uagents-core/uagents_core/contrib/protocols/subscriptions/__init__.py)?

python/src/uagents/experimental/subscription/__init__.py
```diff
@@ -0,0 +1,109 @@
...
+
+        Args:
+            agent_address: The address of the agent making the request.
+
+        Returns:
+            False if the agent is on the free tier and the maximum number of requests
+              has been exceeded, True otherwise.
+        """
+        tier = get_subscription_tier(self._identity, agent_address, self._agentverse)
+        if tier in (TierType.PLUS, TierType.PRO):
```
Dacksus, line 108: simple extension for more demonstrability would be to multiply the `max_requests` for the `PLUS` type and make it unlimited with `PRO`

Dacksus:
> > I wonder if it would make sense to merge it with the general [Subscription protocol](https://github.com/fetchai/uAgents/blob/be012455e3e622c8c4a055e415bd2732898ce362/python/uagents-core/uagents_core/contrib/protocols/subscriptions/__init__.py)?
> 
> The tricky thing and the reason I decided to rename this to `SubscribableProtocol` is that this is meant to wrap any protocol belonging to an agent who has already enabled subscriptions via the `SubscriptionProtocol`, which is more about managing the user subscriptions to agents themselves. But let me know if you thought about this differently.

Yes, I agree that technically they're separate protocols. Conceptually, I feel it would be more convenient for users to only have a single import for everything subscription related, because ultimately you would always need both aspects to make subscriptions work.

Probably at this point the effort is not worth it, but it feels like the right point to integrate would be e.g., at the ChatProtocol level? I.e., make the generic ChatProtocol everyone is using a SubscribableProtocol so that one simply could annotate the message handler for ChatMessages if it's a subscription agent?

PR #742, opened 2025-07-02
feat(uagents-adapter/mcp): add new feature 'ResetTools', validation for arguments and 'session_id'

iliarusskikh:
## Proposed Changes

The ResetTools message type, introduced in the updated protocol.py and handled in adapter.py, is called when a client explicitly sends a ResetTools message to the server. This message is designed to clear session-specific state, such as the message history stored in ctx.storage or any internal state in the mcp_server (assuming that it supports a **reset_state** method).

The ResetTools message is part of the MCP protocol (mcp_protocol_spec). A client would send this message when it wants to reset the state for a specific session (identified by session_id).
For example, a client might send a ResetTools message:
-At the start of a new conversation to ensure no previous tool call parameters or message history are carried over.
-After completing a task to clear any cached data.
-In response to an error or unexpected behavior to reset the server’s state.

The updated adapter.py mitigates the issue by:
-Clearing message history on StartSessionContent.
-Filtering out role: tool messages in the ChatMessage handler.
-Validating tool arguments against inputSchema. These measures reduce the need for frequent ResetTools calls but don’t replace it entirely, as ResetTools provides a client-controlled mechanism for explicit state resets.

Nevertheless, I think that ResetTools may add up complexity on the client-side application. 

## Linked Issues

I have encountered issues when sending user-prompt via ASI1 or ChatProtocol to MCP agent. I have noticed that sometimes the mcp.tools() function would use the arguments from the previous prompt. My assumption was that there is some buffer/resources that are not cleared (cached).

Screenshot 2025-07-02 at 20 18 05

Otherwise, I believe there should be something implemented internally both for function parameter checks and reset_state to simplify the client-side implementation.

There is a chance that I missed something in the current stack implementation that should tackle the issue, FastMCP, or this provided solution/idea is not viable, however, I would highly appreciate the feedback!

## Types of changes

- [ ] Documentation update.

## Checklist

### If applicable

## Further comments

Whether FastMCP class from the mcp.server.fastmcp module has implementation of a reset_state method or not,  this could affect how the ResetTools message is handled in the updated adapter.py and whether additional modifications are needed.

Thanks

### DIFF
--- a/python/uagents-adapter/src/uagents_adapter/mcp/__init__.py
+++ b/python/uagents-adapter/src/uagents_adapter/mcp/__init__.py
@@ -3,7 +3,14 @@
 from importlib import metadata
 
 from .adapter import MCPServerAdapter
-from .protocol import CallTool, CallToolResponse, ListTools, ListToolsResponse
+from .protocol import (
+    CallTool,
+    CallToolResponse,
+    ListTools,
+    ListToolsResponse,
+    ResetTools,
+    ResetToolsResponse,
+)
 
 try:
     __version__ = metadata.version(__package__.split(".")[0])
@@ -19,5 +26,7 @@ __all__ = [
     "ListToolsResponse",
     "CallTool",
     "CallToolResponse",
+    "ResetTools",
+    "ResetToolsResponse",
     "__version__",
 ]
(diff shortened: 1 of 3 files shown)


iliarusskikh:
This issue have been discussed. More tests required with disabling history in mcp adapter library.

PR #753, opened 2025-07-25
feat(core) - Add Helper Utilities to Reduce Agent Chat Protocol Boilerplate in uagents_core

tanaygodse:
## Proposed Changes

Added common chat protocol models, a pre-configured `chat_proto`, and grouped helper methods into a single `ChatObjects` utility class.

## Linked Issues

- Consolidated all chat content models, the `chat_protocol_spec`, and message types (`ChatMessage`, `ChatAcknowledgement`) into `uagents_core/contrib/protocols/chat/__init__.py`.
- Created a single, ready‑to‑use chat_proto instance with a default ACK handler.
- Introduced a new `ChatObjects` class with class‑methods `.text()`, `.end_session()`, and `.resource()` to replace free‑function helpers.
- Removed duplicate helper code from individual agents and updated their imports to use `chat_proto` and `ChatObjects`.

## Types of changes

- [ ] Documentation update.

## Checklist

### If applicable

## Further comments

Grouping all chat‑related utilities into one module and a single helper class reduces code duplication, and makes future extensions (e.g. new content types) straightforward—just add another ChatObjects method and register any protocol handlers in one place.

### DIFF
--- a/python/uagents-core/uagents_core/contrib/protocols/chat/__init__.py
+++ b/python/uagents-core/uagents_core/contrib/protocols/chat/__init__.py
@@ -1,14 +1,20 @@
 """
-This module contains the protocol specification for the agent chat protocol.
+This module contains the protocol specification for the agent chat protocol,
+plus a ready‐to‐use Protocol instance and message‐building helpers.
 """
 
 from datetime import datetime
-from typing import Literal, TypedDict
+from datetime import timezone as _timezone
+from typing import Literal, TypedDict, Union
+from uuid import uuid4
 
 from pydantic.v1 import UUID4
+from uagents import Context, Protocol
 
 from uagents_core.models import Model
 from uagents_core.protocol import ProtocolSpecification
+
+# --- Original content models & spec ---------------------------------------
 
 
 class Metadata(TypedDict):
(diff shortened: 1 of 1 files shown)


python/uagents-core/uagents_core/contrib/protocols/chat/__init__.py
```diff
@@ -122,3 +127,61 @@ class ChatAcknowledgement(Model):
...
 )
+
+
+# --- Helper Class for simplifying Chat Protocol -----------------------------
+
+
+class ChatObjects:
+    """
+    Utility class for constructing ChatMessage objects with common content types.
+    """
```
jrriehl, line 138: It's not clear to me what the value of this class is over standalone utility functions. Maybe it would help if you add an example usage somewhere, maybe in a README?

I have a slight preference for preserving something close to the interface for `create_text_chat` that is used in a lot of agents already (e.g. https://github.com/fetchai/uAgent-Examples/blob/main/6-deployed-agents/geo/google-api-geolocation-agent/chat_proto.py#L22-L30). The reason is that this would make refactoring these agents a bit simpler. But I'd be open to other suggestions if you have a compelling reason or example.

tanaygodse: The main reason for using a class over standalone functions is discoverability and organization. With a class, users can import a single object of `ChatObjects` and easily explore all available methods via IDE auto-completion. This is especially helpful for new users who might not be aware of all the utility functions like `create_resource_chat`. With functions, that discoverability is lost unless they're explicitly mentioned in the documentation being referred to by the user (We would prefer the user to be able to explore the framework intuitively rather than only following examples or documentation).

Additionally, using a class gives us flexibility for the future e.g., we could introduce shared state or configuration (like session_id or some other similar values, etc.) without changing the interface of every method.

Regarding backward compatibility: existing agents using create_text_chat and similar functions inside a `chat_proto.py` will continue to work. But going forward, we’d prefer migrating to this new structure by removing custom `chat_proto.py` files and importing directly from `uagents_core`.

jrriehl: Thanks for the explanation. While I agree with many of your points, my main concern is the potential for confusion when adding a separate class that primarily serves the purpose of adding custom constructors for a subset of the content types. If the main motivation is to reduce boilerplate like setting the type strings, timestamps, and uuids, my proposal would be to add these defaults directly to the constructors of the `ChatMessage` and `Content` objects: https://github.com/fetchai/uAgents/pull/754. If we include this update, would you still see a need to create a separate utility class?

PR #783, opened 2025-09-16
feat(integration): centralize AI address, dedupe port logic, add …

abhifetch:
…HTTP timeouts, fix extras/docs

## Proposed Changes

- Set default AI agent address to agent1qtlpfshtlcxekgrfcpmv7m9zpajuwu7d5jfyachvpa4u3dkt6k0uwwp2lct
- Centralize address resolution via BaseRegisterTool._get_ai_agent_address
- Refactor langchain/crewai adapters to reuse BaseRegisterTool._find_available_port (remove duplicates)
- Add timeout=10s to Agentverse requests (connect/update/register)
- build(pyproject): add extras "langchain" and "crewai"; set Ruff target-version to py310
- docs(README): fix extras install example and import path

## Linked Issues

## Types of changes

- [ ] Documentation update.

## Checklist

### If applicable

## Further comments

### DIFF
--- a/python/uagents-adapter/README.md
+++ b/python/uagents-adapter/README.md
@@ -148,7 +148,7 @@ mcp_adapter.run(agent)
 
 > **Important**: When creating MCP tools, always include detailed docstrings using triple quotes (`"""`) to describe what each tool does, when it should be used, and what parameters it expects. These descriptions are critical for ASI:One to understand when and how to use your tools.
 
-For more detailed instructions and advanced configuration options, see the [MCP Server Adapter Documentation](src/uagents_adapter/mcp/README.md).
+ 
 
 ## A2A Outbound Adapter
 
@@ -205,17 +205,7 @@ The outbound adapter includes payment bridging between AP2 artifacts and Fetch.a
 
 Response parsing prioritizes AP2 data parts in A2A JSON-RPC responses, so carts and payment results propagate as typed objects rather than plain text.
 
-### Environment variables for Skyfire payments
-
-Set these in your environment (or a `.env`) to enable Skyfire token verification/charge flows inside your A2A executors:
-- `SELLER_SERVICE_ID`
-- `SKYFIRE_TOKENS_CHARGE_API_URL` or `SKYFIRE_TOKENS_API_URL`
-- `SELLER_SKYFIRE_API_KEY`
-- `JWKS_URL`
-- `JWT_ISSUER`
-- `SELLER_ACCOUNT_ID`
-- `ASI_API_KEY` (optional; enables LLM routing in `MultiA2AAdapter`)
-
+ 
 ### JSON-RPC message format for A2A
 
 ```json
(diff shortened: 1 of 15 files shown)


python/uagents-adapter/src/uagents_adapter/a2a_outbound/adapter.py
```diff
@@ -162,25 +196,74 @@ async def on_start(ctx: Context):
...
+            ctx.logger.info(f"📩 Received message from {sender}: {type(msg)} object")
+            # For single adapter, just acknowledge rejection
+            await ctx.send(
+                sender,
+                ChatMessage(
+                    timestamp=datetime.now(timezone.utc),
+                    msg_id=uuid4(),
+                    content=[TextContent(type="text", text="Payment rejected.")],
+                ),
+            )
```
gabrielplicks:
```suggestion
            ctx.logger.info(f"📩 Received message from {sender}: {type(msg)} object")
            # We need to convert RejectPayment to DenyCartMandate and send it to the A2A seller agent
            deny_cart_mandate = rejectpayment_to_denycartmandate(msg)
            response = await self._send_to_a2a_agent(
                message={DENY_CART_MANDATE_KEY: deny_cart_mandate.dict()},
                a2a_url=f"http://localhost:{self.a2a_port}",
            )
            # PaymentProtocol round is over, we are out of PaymentProtocol context from here on
            # If we get back a text response, wrap in ChatMessage to continue conversation
            ctx.logger.info(f"🤖 A2A response: {type(response)} object")
            if isinstance(response, str):
                response_msg = ChatMessage(
                    timestamp=datetime.now(timezone.utc),
                    msg_id=uuid4(),
                    content=[TextContent(type="text", text=response)],
                )
            else:
                response_msg = ChatMessage(
                    timestamp=datetime.now(timezone.utc),
                    msg_id=uuid4(),
                    content=[
                        TextContent(
                            type="text",
                            text="❌ Unsupported response type from A2A agent",
                        )
                    ],
                )
            await ctx.send(sender, response_msg)
            ctx.logger.info(
                f"📤 Sent response back to {sender}: {type(response_msg)} object"
            )
```

gabrielplicks: Not sure if it makes sense, but I think here we need to send something to the A2A agent to let it know that the user has rejected the payment, because otherwise we're just sending back a "Payment rejected." message to the buyer (`sender`) but we don't inform the actual A2A seller.

In this code suggestion I send back to the A2A seller a `DENY_CART_MANTATE_KEY` dict containing the rejection message. Then, whatever the A2A agent responds, I just forward as a text message so that the conversation continues. I forward the response because there could be a case in which the seller starts suggesting something else, etc.

python/uagents-adapter/src/uagents_adapter/a2a_outbound/adapter.py
```diff
@@ -435,8 +546,42 @@ async def on_start(ctx: Context):
...
+                    msg_id=uuid4(),
+                    content=[
+                        TextContent(type="text", text="Unexpected response to payment.")
+                    ],
+                )
+            await ctx.send(sender, response_msg)
+
+        @self.payment_proto.on_message(RejectPayment)
+        async def handle_reject_payment(ctx: Context, sender: str, msg: RejectPayment):
+            self._last_agent_for_sender.pop(sender, None)
```
gabrielplicks, line 651: Should we do the same as in the Single adapter here? We pop the sender because we are not tracking it any more mandates from it, but we could inform the A2A seller about the rejection as suggested for the Single adapter.

abhifetch: sure , we can do this for single adapter as well.

python/uagents-adapter/src/uagents_adapter/a2a_outbound/adapter.py
abhifetch: Yes—good catch. We should notify the A2A seller on rejection.
totally missed it.

python/uagents-adapter/README.md
```diff
@@ -195,12 +195,61 @@ if __name__ == "__main__":
...
+lsof -ti TCP:9999,8220,10020 | xargs -I {} kill -9 {} || true
+cd "/Users/abhi/Desktop/a2p protocol/experiment-ap2/a2a-seller-agent"
+PYTHONPATH="/Users/abhi/Desktop/a2p protocol/uAgents/python/uagents-adapter/src:$PYTHONPATH" python3 av_adapter.py
+```
 
+Multi-server (coordinator + specialists):
+```bash
+lsof -ti TCP:8200,10020,10022,10023 | xargs -I {} kill -9 {} || true
+cd "/Users/abhi/Desktop/a2p protocol/experiment-ap2/a2a-multi-server"
+PYTHONPATH="/Users/abhi/Desktop/a2p protocol/uAgents/python/uagents-adapter/src:$PYTHONPATH" python3 run_multi_adapter.py
```
jrriehl: You've got some local paths here that you probably want to update.

python/uagents-adapter/src/uagents_adapter/__init__.py
```diff
@@ -2,13 +2,31 @@
...
+
+    A2A_AVAILABLE = True
+except ImportError:
+    A2A_AVAILABLE = False
+    # Create dummy classes for when A2A is not available
+    A2ARegisterTool = None
+    A2AAgentConfig = None
+    MultiA2AAdapter = None
+    SingleA2AAdapter = None
+    a2a_servers = None
```
jrriehl: Is this really needed? The conditional import at the top level seems okay, but if another module tries to import one of these when it is not available, I would rather see an `ImportError` raised.

PR #785, opened 2025-09-22
feat(uagents): set default network to mainnet

jrriehl:
## Proposed Changes

Set default network to mainnet and update logging messages accordingly.

All agents will still register by default on the Almanac API, but we switch the default network to mainnet. To register on testnet instead, agents should set the network to `testnet` explicitly with:
```python
agent = Agent(..., network="testnet")
```

## Types of changes

- [ ] Documentation update.

## Checklist

### If applicable

### DIFF
--- a/python/poetry.lock
+++ b/python/poetry.lock
@@ -2254,14 +2254,14 @@ files = [
 
 [[package]]
 name = "uagents-core"
-version = "0.3.7"
+version = "0.3.10"
 description = "Core components for agent based systems"
 optional = false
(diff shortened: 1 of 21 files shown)


python/src/uagents/resolver.py
```diff
@@ -68,23 +68,6 @@ def is_valid_prefix(prefix: str) -> bool:
     return prefix in valid_prefixes
 
 
-def parse_prefix(prefix: str) -> AgentNetwork:
```
jrriehl, line 71: Not used anywhere.

PR #837, opened 2026-02-25
feat(uagents): support zero-downtime deployment

jrriehl:
## Proposed Changes

Gracefully shutdown all agent tasks in the correct order to support zero-downtime agent deployment.

Most importantly, the message receivers stop processing new messages but in-progress handlers are allowed to finish up to a configurable timeout. During this time, the readiness probe will fail and a new agent process can be started to immediately start receiving new messages.

When using a rolling deployment service like kubernetes, an additional parameter `mark_inactive_on_shutdown` should be set to `False` because it is safe to assume that another agent process will take the place of the shutting down agent and therefore the agent will remain active.

## Linked Issues

## Types of changes

- [ ] Documentation update.

## Checklist

### If applicable

### DIFF
--- a/python/src/uagents/communication.py
+++ b/python/src/uagents/communication.py
@@ -102,8 +102,7 @@ class Dispenser:
                 except asyncio.QueueEmpty:
                     break
                 except Exception as ex:
-                    # Log error but continue draining to prevent message loss
-                    LOGGER.error(f"Error processing envelope during shutdown: {ex}")
+                    LOGGER.exception(f"Error processing envelope during shutdown: {ex}")
 
             LOGGER.info("Shutting down dispenser...complete")
 


python/src/uagents/agent.py
```diff
@@ -399,7 +404,15 @@ def __init__(
...
         self._message_tasks: set[asyncio.Task] = set()
+        self._interval_tasks: set[asyncio.Task] = set()
+        self._rest_tasks: set[asyncio.Task] = set()
+        self._registration_task: asyncio.Task | None = None
+        self._dispenser_task: asyncio.Task | None = None
+        self._message_queue_task: asyncio.Task | None = None
         self._handle_messages_concurrently = handle_messages_concurrently
+        self._shutdown_timeout = shutdown_timeout
+        self._mark_inactive_on_shutdown = mark_inactive_on_shutdown
+        self._shutting_down = False
```
zmezei: Is this flag used anywhere? I see that it's set to True during shut down but I don't see it being used anywhere.

zmezei: Should it be used in the liveness probe of the agent?

jrriehl: Good catch, I removed another one of these from the asgo server, but indeed this one is also no longer used.

python/src/uagents/asgi.py
```diff
@@ -137,6 +137,11 @@ async def _asgi_send(
 
     async def handle_readiness_probe(self, headers: CaseInsensitiveDict, send):
         """Handle a readiness probe sent via the HEAD method."""
+        # Fail readiness probe immediately during shutdown
+        if self._server and self._server.should_exit:
```
zmezei, line 142: What does set this `self._server.should_exit` to True?

python/src/uagents/agent.py
```diff
@@ -1168,6 +1192,63 @@ async def _shutdown(self):
...
                 self._logger.exception(f"Exception in shutdown handler: {ex}")
 
+    async def _shutdown(self, tasks: list[asyncio.Task]):
+        """Perform graceful agent shutdown."""
+        # Step 1: Stop the registration loop
+        if self._registration_task:
+            self._registration_task.cancel()
+            await asyncio.gather(self._registration_task, return_exceptions=True)
+
+        # Step 1: Register as inactive in the Almanac (if configured)
```
zmezei: There is already a Step 1. above

python/src/uagents/asgi.py
jrriehl: It starts `True` and then set to `False` when intercepting the SIGTERM:
```python
    def handle_exit(self, sig: int, frame: FrameType | None) -> None:
        self._captured_signals.append(sig)
        if self.should_exit and sig == signal.SIGINT:
            self.force_exit = True  # pragma: full coverage
        else:
            self.should_exit = True
```

zmezei: Oh okay, so it will be set by for example kubernetes or by ctrl+c when running locally, I see.

zmezei (approved):

python/src/uagents/agent.py
```diff
@@ -1211,14 +1294,16 @@ async def run_startup_tasks(self):
...
         for func, period in self._interval_handlers:
-            self._loop.create_task(
+            task = self._loop.create_task(
                 _run_interval(func, self._logger, self._build_context, period)
             )
+            self._interval_tasks.add(task)
+            task.add_done_callback(self._interval_tasks.discard)
 
     def start_message_receivers(self):
         """Start message receiving tasks for the agent."""
```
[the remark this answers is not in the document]
jrriehl: Resolved

PR #863, opened 2026-04-01
feat(core): adapters process chat message in a background task + initial error reporting

lrahmani:


python/uagents-core/uagents_core/adapters/a2a/agentverse_sdk.py
```diff
@@ -170,6 +172,16 @@ async def _chat(self, request: Request) -> Response:
         if len(msg.content) == 1 and isinstance(msg.content[0], StartSessionContent):
             return JSONResponse({})
 
+        return JSONResponse(
+            {},
+            background=BackgroundTask(
+                self._process_chat_message, request=request, env=env, msg=msg
+            ),
+        )
```
jrriehl, line 195: I presume this supports asynchronous messaging only. Should we at least handle the header `x-uagents-connection: sync` with a error message?

lrahmani: Yes, I am not supporting async messaging yet

PR #868, opened 2026-04-10
fix(a2a-adapter): add follow_redirects=True to httpx AsyncClient calls

bleib1dj:
## Summary

`httpx.AsyncClient` defaults `follow_redirects` to `False`, causing 308 Permanent Redirect responses from Agentverse and A2A endpoints to be returned as errors instead of being followed. This affects agent discovery, health checks, message sending, and LLM routing in the A2A outbound adapter.

This is a companion fix to [flockx-official/platform#11263](https://github.com/flockx-official/platform/pull/11263), which addresses the same issue in the platform codebase.

## Notable changes

- **`a2a_outbound/adapter.py`** - 4 `httpx.AsyncClient` instantiations now include `follow_redirects=True`:
  - `_send_to_a2a_agent` (message sending)
  - `_discover_and_health_check_agents` (agent card fetching)
  - `route_to_llm` (LLM API routing)
  - Agent card well-known endpoint fetching

## How tested

- Code review of all `httpx.AsyncClient` instantiations in the adapter
- Confirmed the change is additive - no behavioral change for endpoints that do not redirect

## Risks / breaking changes

None. Adding `follow_redirects=True` makes httpx behave like `requests` and `aiohttp` do by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,20 +72,19 @@ You should see the results in your terminal.
 
 ## 📖 Documentation
 
-Please see the [official documentation](https://fetch.ai/docs) for full setup instructions and advanced features.
+Please see the [official documentation](https://uagents.fetch.ai/docs) for full setup instructions and advanced features.
 
-- [👋 Introduction](https://fetch.ai/docs/concepts)
-- [💻 Installation](https://fetch.ai/docs/guides/agents/installing-uagent)
+- [👋 Introduction](https://uagents.fetch.ai/docs)
+- [💻 Installation](https://uagents.fetch.ai/docs/getting-started/install)
 - Tutorials
-  - [🤖 Create an agent](https://fetch.ai/docs/guides/agents/create-a-uagent)
-  - [🛣️ Agent Communication](https://fetch.ai/docs/guides/agents/communicating-with-other-agents)
-  - [🍽️ Restaurant Booking Demo](https://fetch.ai/docs/guides/agents/booking-demo)
+  - [🤖 Create an Agent](https://uagents.fetch.ai/docs/getting-started/create)
+  - [🛣️ Agent Communication](https://uagents.fetch.ai/docs/guides/communication)
+  - [🍽️ ASI:One Compatible Agent](https://uagents.fetch.ai/docs/examples/asi-1)
 - Key Concepts:
-  - [📍Addresses](https://fetch.ai/docs/guides/agents/getting-uagent-address)
-  - [💾 Storage](https://fetch.ai/docs/guides/agents/storage-function)
-  - [📝 Interval Tasks](https://fetch.ai/docs/guides/agents/interval-task)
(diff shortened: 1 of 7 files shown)


Warren-Bertram:
## Real-world repro context: hosted mailbox A2A agent (PropertyAI Advisor)

Sharing an external A2A agent setup that hits this exact class of failure. Happy to help validate once this lands.

### Setup / workflow

1. Built a Starlette A2A agent (`a2a-sdk` 1.x) with JSON-RPC at `/` and agent card at `/.well-known/agent-card.json`.
2. Connected it in Agentverse via **Launch an Agent → Connect Agent → A2A Protocol → Mailbox relay** (not public endpoint).
3. Deployed always-on on Railway with Agentverse bridge + mailbox enabled (`USE_MAILBOX=1`), polling the mailbox with the URI from onboarding (`handle:secret@agentverse.ai/...`).
4. Public JSON-RPC on the Railway URL answers correctly when called directly (e.g. `hello` / `SendMessage` with `A2A-Version: 1.0`).
5. Chat from Agentverse / ASI:One does **not** produce any incoming message on the Railway poller — no `[incoming message]` logs.

Agent listing: mailbox-relay A2A agent **property-ai-advisor** / **Property AI Advisor**.  
Public endpoint (works when called directly): `https://property-ai-agent-production.up.railway.app/`

### What fails

When chatting to the hosted A2A agent through Agentverse/ASI:One we see **HTTP 308 Permanent Redirect** failures on the client side.

Root cause matches this PR: `httpx.AsyncClient` defaults to `follow_redirects=False`, so 308s from Agentverse/A2A endpoints (often trailing-slash redirects from Starlette / slash-sensitive paths) are returned as errors instead of being followed.

That lines up with:

- Starlette (and many servers) answering slash mismatches with **308**
- Agentverse/A2A clients using httpx with `follow_redirects=False`
- Discovery / registration evaluation / chat failing before mailbox handoff completes → **no inbound message on the Railway poller**

### Ask (same as this PR)

Please set `follow_redirects=True` on the httpx clients used for:

- sending messages to A2A agents
- agent-card / well-known discovery
- health checks / registration evaluation

Same class of fix is needed anywhere Agentverse/ASI:One calls A2A over httpx (companion: flockx-official/platform#11263).

### Why this matters beyond this agent

Most external A2A setups hit this: JSON-RPC served at `/` (or any slash-sensitive path) + trailing-slash agent-card URLs. Following redirects on the platform/client side is the proper fix; agent deploys can stay as-is meanwhile.

### Extra constraint we hit

URI was fixed at mailbox onboarding; we cannot re-evaluate or switch to a public endpoint without a new profile. So until redirect-following lands, mailbox chat stays broken even though the public Railway JSON-RPC endpoint is healthy.

Thanks for the fix — landing this should unblock mailbox-relay A2A agents like this one.

PR #876, opened 2026-04-17
fix(uagents): handle intermediate responses in send_and_receive

### LINKED ISSUE #798: Fix with ctx.send_and_receive.
### Prerequisites

- [x] I checked the [documentation](https://fetch.ai/docs) and found no answer to my problem
- [x] I checked the [existing issues](https://github.com/fetchai/uAgents/issues) and made sure there are no similar bug reports

### Category

Bug (unexpected behavior)

### Expected Behavior

While getting a message back in send_and_recieve from chat protocol. it should ignore chatAcknowledgment and consider only chatMessage

### Observed Behavior

I noticed a bug from hackathon participants where if an agent receives a ChatAcknowledgement message before a ChatMessage, and in the case ctx.send_and_receive expects a ChatMessage, the ctx.send_and_receive function errors out. It seems to detect the first message type and fails to process the subsequent one properly. This could happen to other message schemas too other than ChatProtocol-related ones.

### To Reproduce

_No response_

### Version

0.22.10

### Environment Details (Optional)

```markdown

```

### Failure Logs (Optional)

```shell

```

### Additional Information (Optional)

```markdown

```

pshkv:
## Summary
- keep pending responses in arrival order instead of treating the first response as terminal
- let `ctx.send_and_receive()` ignore intermediate unexpected response types while continuing to wait within the same timeout window
- preserve the existing failure contract when only unexpected response types arrive
- add a regression test for the acknowledgement-then-message pattern

## Why
This fixes issue #798, where `ctx.send_and_receive()` can fail if an acknowledgement or other unexpected message arrives before the expected response type.

## Testing
- `python -m pytest -q python/tests/test_context.py -k send_and_receive`
- `python -m compileall python/src/uagents python/tests/test_context.py`

Closes #798

### DIFF
--- a/python/src/uagents/context.py
+++ b/python/src/uagents/context.py
@@ -5,7 +5,7 @@ import logging
 import uuid
 from abc import ABC, abstractmethod
 from collections.abc import Callable
-from time import time
+from time import monotonic, time
 from typing import TYPE_CHECKING, Any
 
 import requests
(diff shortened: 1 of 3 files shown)


jrriehl:
We went with a slightly different approach on this fix, but thanks for your contribution and please let us know if you still have issues: https://github.com/fetchai/uAgents/pull/882.

PR #930, opened 2026-08-12
feat(experimental): stream ChatAgent replies

Alejandro-Morales:
## Proposed Changes

- Stream ChatAgent plain-text replies as StartStream → TextContent deltas → EndStream instead of waiting for a full completion
- Keep tool selection non-streaming; stream only the final user-facing reply after tools run
- Include the assistant tool_calls message in the post-tool LLM follow-up so providers accept the tool result

## Types of changes

- [ ] Documentation update.

## Checklist

### DIFF
--- a/python/src/uagents/experimental/chat_agent/llm.py
+++ b/python/src/uagents/experimental/chat_agent/llm.py
@@ -87,7 +87,7 @@ class LLMConfig(BaseModel):
     url: str
     parameters: LLMParams
     api_key: str | None = None
-    stream: bool = True # Set False for providers that do not support stream
+    stream: bool = True  # Set False for providers that do not support stream
 
     @classmethod
     def asi1(cls, model: str = DEFAULT_ASI1_MODEL, stream: bool = True) -> "LLMConfig":
--- a/python/src/uagents/experimental/chat_agent/llm.py
+++ b/python/src/uagents/experimental/chat_agent/llm.py
@@ -87,9 +87,10 @@ class LLMConfig(BaseModel):
     url: str
     parameters: LLMParams
     api_key: str | None = None
+    stream: bool = True # Set False for providers that do not support stream
 
     @classmethod
-    def asi1(cls, model: str = DEFAULT_ASI1_MODEL) -> "LLMConfig":
+    def asi1(cls, model: str = DEFAULT_ASI1_MODEL, stream: bool = True) -> "LLMConfig":
         api_key = os.getenv("ASI1_API_KEY")
         if api_key is None:
             raise ValueError("Please set ASI1_API_KEY environment variable.")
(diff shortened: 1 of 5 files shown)


python/src/uagents/experimental/chat_agent/llm.py
```diff
@@ -209,6 +207,28 @@ async def process(
 
         raise RuntimeError("LLM returned neither tool_calls nor content.")
 
+    async def process_stream(self, messages: list[dict]) -> AsyncIterator[str]:
+        """Yield plain-text content deltas from a streaming completion."""
+        kwargs = self._get_base_kwargs(
+            messages, exclude_params={"system_prompt", "tool_choice"}
+        )
+        kwargs["stream"] = True
+        kwargs.pop("parallel_tool_calls", None)
```
jrriehl, line 218: Since `stream` is available in LLMConfig, would it not make more sense to get it from there in `_get_base_kwargs`? Also, are these settings for `system_prompt`, `tool_choice`, and `parallel_tool_calls` specific to streaming? If not, maybe they could also be handles in `_get_base_kwargs`?

python/src/uagents/experimental/chat_agent/protocol.py
```diff
@@ -143,8 +146,32 @@ async def _chat_handler(ctx: Context, sender: str, msg: ChatMessage):
...
+                    return await self.send_stream(
+                        ctx,
+                        sender,
+                        [
+                            {
+                                "role": "system",
+                                "content": self._llm._system_content([]),
+                            },
+                            *messages,
+                        ],
```
jrriehl, line 161: Why is the system content only added when `self._tools` is populated?

jrriehl: I guess it's because the system content is added inside `process`, but is there a particular reason for doing differently between streaming and non-streaming?

jrriehl (commented):
Looking good! I added a couple of questions.

python/src/uagents/experimental/chat_agent/protocol.py
Alejandro-Morales: Yess process() already adds the system message. complete and process_stream dont, so we add it here on the no tools path

python/src/uagents/experimental/chat_agent/llm.py
Alejandro-Morales: That helper is shared with process(), that must stay non streaming for tool selection. LLMConfig.stream only decides whether we call process_stream for the user reply and once this is running so we can hardcode kwargs["stream"] = True, 

no, system_prompt, tool_choice, and parallel_tool_calls arent streaming specific. I could move system_prompt  into _get_base_kwargs if you think would be best, but tool_choice  and parallel_tool_calls still need to stay for tool calls in process()

